PR #1, opened 2019-06-13
Minor readme changes

JakeChvatal:
Suggested some minor changes to the readme and code style. If you have other suggested style changes for the repository, a cohesive organization strategy should be established before moving forward.

It is worth noting that some work may have to be done to format existing code, especially in the parse_audit.js file - I will handle this. We should look into renaming and/or combining files here as well, as having multiple files with similar 'parser' names is confusing.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -62,5 +62,9 @@ typings/
 
 # degree audit
 Web Audit.html
+
+Web\ Audit.html
+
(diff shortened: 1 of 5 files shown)


bugsalexander:
Closed - waiting to use class system with methods until it is functional.

PR #9, opened 2019-06-19
fix-parser

JakeChvatal:
Moved logic for different parts of the degree audit into individual functions.
Fixed the logic for producing the courses that are required.

To test:
Run the code as per the README of this repository.
Ensure that the degree audit produced is as desired.

### DIFF
--- a/html_parser.js
+++ b/html_parser.js
@@ -1,144 +1,201 @@
+/**
+ * HTML Degree Audit parser built for Northeastern University's degree audit.
+ * By Jacob Chvatal for Northeastern Sandbox.
+ *
+ * Filters through a Northeastern degree audit, extracting its relevant information
+ * to an easy to work with JSON file tracking major, graduation date,
+ * classes and NUPaths taken and in progress as well as requirements to take.
+ */
+
 const fs = require('fs');
 
-const INPUT = './Web\ Audit.html';
+// location of the input file
+const INPUT = '../Web\ Audit.html';
+
+// location of the output file
 const OUTPUT = 'parsed_audit.json';
 
-fs.readFile(INPUT, 'utf8', (err, data) => {
-	if(err) {throw err;}
-	// else, process the data
+/**
+ * Gets the major associated with this degree audit.
+ * @param json  The json file to which the major should be added.
+ * @param line  The line at which the major can be found.
+ */
+function identify_major(json, line) {
+    json.data.major = line.substring(line.search('>') + 1, line.search(' - Major'));
+}
 
-	json = { 
-		completed: {
-			classes:[],
-			nupaths:[]
-		},
-		inprogress: {
-			classes:[],
-			nupaths:[]
-		},
-		requirements: {
-			classes:[],
-			nupaths:[]
-		},
-		data: {
-		
-		}
-	};
+/**
+ * Gets the year this degree audit was created.
+ * @param json  The json file to which the major should be added.
+ * @param line  The line containing the year.
+ */
+function get_year(json, line) {
+    json.data.year = line.substring(line.search('CATALOG YEAR:') + 'CATALOG YEAR: '.length, line.search('CATALOG YEAR:') + 'CATALOG YEAR: '.length + 4);
+}
 
-	var lines = data.split('\n');
-	for(var i = 0; i < lines.length; i++) {
+/**
+ * Gets the expected graduation date of the degree audit.
+ * @param json  The json file to which this graduation date should be added.
+ * @param line  The line which contains the graduation date desired.
+ */
+function get_grad_date(json, line) {
+    json.data.grad = line.substring(line.search('GRADUATION DATE: ') + 'GRADUATION DATE: '.length, line.search('GRADUATION DATE: ') + 'GRADUATION DATE:  '.length + 7);
+}
 
-		// finds major
-		if(contains(lines[i], 'Major')) {
-			json.data.major = lines[i].substring(lines[i].search('>') + 1, lines[i].search(' - Major'));
-		}
-		// finds year
-		if(contains(lines[i], 'CATALOG YEAR')) {
-			json.data.year = lines[i].substring(lines[i].search('CATALOG YEAR:') + 'CATALOG YEAR: '.length, lines[i].search('CATALOG YEAR:') + 'CATALOG YEAR: '.length + 4);
-		}
+/**
+ * Gets the NUPaths associated with this degree audit.
+ * @param json  The json file to which the NUPaths should be added.
+ * @param lines The lines to be scanned for the NUPaths.
+ * @param i     The index of the line at which we currently stand.
+ */
+function get_nupaths(json, lines, i) {
+    i++;
+    while(contains(lines[i], '<br>')) {
+        var toAdd = lines[i].substring(lines[i].indexOf("(") + 1, lines[i].indexOf("(") + 3)
+        // omits bad stuff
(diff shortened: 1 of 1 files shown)


JakeChvatal:
Note: the parser currently prints out its analysis of courses to take before exporting the file to make validating this PR easier. Ensure that any console.log statements are removed before the PR is accepted.

html_parser.js
```diff
@@ -1,144 +1,201 @@
...
+}
+
+/**
+ * Adds the courses taken so far to the current JSON file.
+ * @param json  The JSON file to which the courses to be taken should be added.
+ * @param line  The line which contains the course taken.
+ */
+function add_course_taken(json, line) {
+    var course = {};
+    var courseString = line.substring(line.search('(FL|SP|S1|S2)'));
```
bugsalexander: var instead of let or const?

html_parser.js
```diff
@@ -67,35 +67,61 @@ function get_nupaths(json, lines, i) {
         }
         i++;
     }
+}
 
+/**
+ * Determines whether a function contains a number.
+ * @param n  The String which could contain a number.
+ * https://stackoverflow.com/questions/5778020/check-whether-an-input-string-contains-a-number-in-javascript#28813213
```
bugsalexander: Should I also leave links to stackoverflow copy/pasted code?

html_parser.js
```diff
@@ -91,7 +91,7 @@ function add_courses_to_take(json, lines, j) {
     for(var i = 0; i < courseList.length; i++) {
         var course = { };
         if(!(courseList[i] == ' ' || courseList[i] == '' || courseList[i] == '****')) {
-            course.attr = courseList[i].substring(0, 4).split(' ').join('');
+            course.course.subject = courseList[i].substring(0, 4).split(' ').join('');
```
bugsalexander: I think this should be course.subject instead of course.course.subject

bugsalexander:
Additionally, [here](https://github.com/sandboxneu/scheduleneu/blob/9b27a1cbcd210ff7f62f25dd2087fb75e741db4a/html_parser.js#L144) (lines 143 and 144), season and year should be replaced with a new metric of when classes were taken, "termId", which looks something like "201830" (in String format) meaning 2018 => 2018, 30 => spring.

Also note that each termId starts in the fall of the previous year. For example the termIds starting with 2018: 

* 201810 = fall 2017
* 201830 = spring 2018
* 201840 = summerI 2018
* 201850 = summer full 2018
* 201860 = summerII 2018

html_parser.js
```diff
@@ -1,144 +1,201 @@
...
+    course.hon = contains(line, '\(HON\)');
+
+    // ap courses that do not count for credit do not have numbers / attributes for corresponding college courses
+    if(!contains(courseString, 'NO AP')) {
+        course.attr = line.substring(line.search('(FL|SP|S1|S2)') + 5, line.search('(FL|SP|S1|S2)') + 9).replace(' ', '');
+        course.number = courseString.substring(9, 14);
+    }
+    course.credithours = courseString.substring(18, 22);
+    course.season = new RegExp('(FL|SP|S1|S2)').exec(line)[0];
+    course.year = line.substring(line.search('(FL|SP|S1|S2)') + 2, line.search('(FL|SP|S1|S2)') + 4);
```
bugsalexander: Here's the spot I was referencing in the comment above.

html_parser.js
JakeChvatal: I remember telling you they should be used instead of var awhile back... rookie mistake of me.

html_parser.js
JakeChvatal: I've seen others do this in their production JavaScript code; I like it as it's more transparent and respects the original author, and without it, if copied code is identified, the implication is that if some of the code has been stolen without citation any of it could be. Always cite sources!

html_parser.js
JakeChvatal: > Additionally, [here](https://github.com/sandboxneu/scheduleneu/blob/9b27a1cbcd210ff7f62f25dd2087fb75e741db4a/html_parser.js#L144) (lines 143 and 144), season and year should be replaced with a new metric of when classes were taken, "termId", which looks something like "201830" (in String format) meaning 2018 => 2018, 30 => spring.
> 
> Also note that each termId starts in the fall of the previous year. For example the termIds starting with 2018:
> 
> * 201810 = fall 2017
> * 201830 = spring 2018
> * 201840 = summerI 2018
> * 201850 = summer full 2018
> * 201860 = summerII 2018
Additionally, [here](https://github.com/sandboxneu/scheduleneu/blob/9b27a1cbcd210ff7f62f25dd2087fb75e741db4a/html_parser.js#L144) (lines 143 and 144), season and year should be replaced with a new metric of when classes were taken, "termId", which looks something like "201830" (in String format) meaning 2018 => 2018, 30 => spring.

Also note that each termId starts in the fall of the previous year. For example the termIds starting with 2018:

* 201810 = fall 2017
* 201830 = spring 2018
* 201840 = summerI 2018
* 201850 = summer full 2018
* 201860 = summerII 2018

Got it - I had lopped off the two numbers at the end as I thought the year would be easier to work with. Good catch.

JakeChvatal:
I didn't deliberately set the pipeline to 'Review/QA', I pushed intermediate changes. Sorry if that was confusing or if you were notified. I'll let you know when this is actually to be reviewed.

html_parser.js
```diff
@@ -147,21 +154,48 @@ function add_course_taken(json, line) {
...
+        case "FL":
+            return termid + "40";
+        case "SP":
+            return termid + "10";
+        case "S1":
+            return termid + "20";
+        case "S2":
+            return termid + "30";
+        default:
+            throw "The given season was not a member of the enumeration required."
```
bugsalexander: The numbers here are incorrect. Additionally, this switch-case is missing a case, summer full. See my earlier comment here for details: 

https://github.com/sandboxneu/scheduleneu/pull/9#issuecomment-504398461

bugsalexander: Some of the documentation can be found [here](https://github.com/ryanhugh/searchneu/blob/master/docs/API.md). 

Additionally I'm not sure whether the "year" of a class in the HTML audit is the same "year" as used in the SearchNEU representation (ex. 2019 starts in fall 2018, meaning fall 2018 semester's "year" is actually 2019).

html_parser.js
```diff
@@ -174,7 +174,7 @@ function add_course_taken(json, line) {
 
 /**
  * Determines the termId parameter of a course with the season and year of the course.
- * @param {String} season   The season of the course ('FL', 'SP', 'S1', 'S2')
+ * @param {String} season   The season of the course ('FL', 'SP', 'S1', 'S2', 'SF')
```
bugsalexander: Here you say "SF" = summer full...

html_parser.js
```diff
@@ -184,14 +184,19 @@ function get_termid(season, year) {
...
-        case "S1":
-            return termid + "20";
-        case "S2":
+        case "SP": // Spring term
             return termid + "30";
+        case "S1": // Summer 1 term
+            return termid + "40";
+        case "S2": // Summer 2 term
+            return termid + "60";
+	case "SM": // Full Summer term (typically reserved for graduate courses)
```
bugsalexander: ... but here you say "SM" = summer full. Since these are values used by the HTML audit itself, they should be decided by looking at the HTML. 

If we need someone who has taken summer full courses, I know Max Daniels is taking summer full courses.

bugsalexander: update: I just emailed Max. I'll slack you when he responds.

PR #10, opened 2019-06-20
Course requirements

bugsalexander:
Added ability to query course requirements from SearchNEU JSON data. Automatically downloads SearchNEU JSON fall/spring data files from public endpoints.


JakeChvatal:
Looks great!

PR #14, opened 2019-07-22
Fix parser

JakeChvatal:
The parser is now able to perform all of the desired and described functionality with the two sample degree audits utilized. 

To test: ```yarn test```

Please feel free to look at the test suite and contribute as you see fit - what is needed most are further degree audits to be verified, which can be added to the folder with the sample degree audits, imported as constants at the top of the file and added to the array to verify their structure. I believe coverage is comprehensive, though the form of the courses required is not comprehensively verified as of yet.

This is because I wanted to verify that this is the form we are sticking with for the various forms of required courses; if any other structure, names, etc would be more useful, those should be incorporated instead.


dajinchu:
did you forget to commit `package.json`

bugsalexander:
> did you forget to commit `package.json`

I messaged him about it earlier, he said he'll add it when he gets home

PR #16, opened 2019-07-23
fix - Normalize HTML parser over different audits

bugsalexander:


JakeChvatal:
This PR should be accepted regardless of representational issues; those will be covered in different PRs. This one is ready to go!

JakeChvatal:
The audit has not yet been normalized, but it has changed representation, which I feel is another PR. Changing representation of data should be prioritized over conforming the code to additional audits.

PR #17, opened 2019-07-23
addition - basic schedule production

bugsalexander:


bugsalexander:
Whoops, accidentally requested a review. Still need to implement topological sort/coffman-graham's!

bugsalexander:
I can't read and apparently you need to reverse all the edges of the graph before running Coffman/Graham's algorithm  –__–  still have that to do.

PR #20, opened 2019-07-24
Added names as unique identifiers to courses

JakeChvatal:
As discussed in the issue.

To test: ```yarn test```

Also ensure that the produced audit visible in the test case uses the desired convention for names and that the 'no duplicates' heuristic is now properly respected.


test/html_parser.test.js
```diff
@@ -710,6 +713,365 @@ test('Verifies that the CS degree audit is properly reproduced by the code', ()
...
+                        classId: '4500' 
+                    },
+                    { 
+                        subject: 'THTR', 
+                        classId: '1170' 
+                    },
+                    { 
+                        subject: 'CS', 
+                        list: [ '4100', '4300', '4410','4150','4550','4991','4900'],
+                        num_required: '1'
```
bugsalexander: It would be cool if we had a "type" property here for the type of requirement needed. I'll open another issue. 

Ex. this one might have type "or" or "union", for ranges we could have type "range"

test/html_parser.test.js
```diff
@@ -1225,6 +1339,7 @@ test('Verifies that the CS Math degree audit is properly reproduced by the code'
                         "hon":false,
                         "subject":"ARTS",
                         "classId":"1990",
+                        "name": "Ceramics:WheelThrowing",
```
bugsalexander: Wheel Throwing LOL

JakeChvatal: There's a harry potter course!

test/html_parser.test.js
JakeChvatal: I have another issue out for that - figure it should have been distinct.

PR #26, opened 2019-08-19
Fixed code for ease of testing

bugsalexander:
Lots of general fixes to improve code architecture, both for testing, and for eventual web-app.

* updated jest version in package.json and yarn.lock
* refactored json_parser.js to export main function for testing
* separated loading of files to json_loader.js, also exported for testing
* exported Graph class from json_parser.js for testing
* added testing for topological sort in Graph class.
* added async testing for json_parser.js

The main fix was separating the file loading from schedule building. This way, the server would only need to load the files once and store them to pass them to future schedule builds, rather than inherently loading the files every single time a schedule is built. This has the added side effect of making the tests not take time proportional to the number of schedules built, but rather to the number of times the files are loaded (once, ever). 

If the tests timeout on the first try, it's probably because json_loader had to re-download the classmaps from SearchNEU. It should work on the second try.


JakeChvatal:
Looks great!

PR #31, opened 2019-09-22
added prettier.rc and formatted all files

NilsBacke:
This should keep our styling options the same between us. Make sure to install and enable the prettier plugin for VScode for this to take effect.

No content in the files was changed, just the formatting.

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,7 @@
+{
+  "trailingComma": "es5",
+  "singleQuote": false,
+  "semi": true,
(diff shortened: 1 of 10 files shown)


NilsBacke:
Anyone know why the tests are failing?

bugsalexander:
> Anyone know why the tests are failing?

Yeah, it's because SearchNEU is no longer hosting the course data for last spring and fall, and so the files it downloads are getting parse errors because they aren't json. I'll try to have a fix by tomorrow's meeting.

PR #32, opened 2019-09-22
Missing classmaps

bugsalexander:
Now safe-deletes invalid json files downloaded from SearchNEU (for example, if they're no longer being hosted :P)

I also removed a file that should never have made it to the actual repo, sorry!


src/json_loader.ts
```diff
@@ -2,37 +2,46 @@
...
+}
 
 /**
  * Provides an array of the links to the classMap files for a specified year.
  * @param year The 4-digit year to retrieve links for
  * @returns The strings of the links for the five files.
  */
 const getClassMapLinks = (year: number): string[] =>
-SEASONS.map((season) => "https://searchneu.com/data/v2/getTermDump/neu.edu/" + year + season + ".json");
+SEASONS.map((season) => getClassMapLinkForTermId((year * 100) + season));
```
yasz24: Maybe add a comment, clarifying why you need to do the `(year * 100) + season)`

src/json_loader.ts
```diff
@@ -2,37 +2,46 @@
...
 
 /**
  * Provides an array of filepath locations to the classMap files based on a provided year.
  * @param year The target year. expected as a string or number. in the form "2019".
  * @returns The names of the five files.
  */
-const getClassMapFilePaths = (year: number): string[] => SEASONS.map((season) => "./" + year + season + ".json");
-
-const SEASON_PATHS: string[] = SEASONS.map((season) => "./" + YEAR + season + ".json");
+const getClassMapFilePaths = (year: number): string[] => SEASONS.map((season) => getClassMapLinkLocal((year * 100) + season));
```
yasz24: This minor, but same as above.

PR #34, opened 2019-10-06
Data def

bugsalexander:
Added new data definitions for Schedule and Major.
Added simple warning system, that takes in a Schedule.
Added a function to convert from IInitialScheduleRep to Schedule (old schedule data def to new). 
Added tests for simple warning system.


src/types.ts
```diff
@@ -288,3 +288,130 @@ export interface IScheduleCourse {
...
+    season: Season;
+    year: number;
+    termId: number;
+    id: number;
+    status: Status;
+    classes: ScheduleCourse[];
+}
+
+/**
+ * A Statis is one of on CO-OP, CLASSES, or INACTIVE
```
NilsBacke: Typo :)

PR #38, opened 2019-10-13
type altercations

NilsBacke:


backend/src/types.ts
```diff
@@ -267,19 +267,30 @@ export interface INEUPrereqCourse {
  * @param scheduled The scheduled courses.
  */
 export interface ISchedule {
-    completed: ICompleteCourse[];
-    scheduled: string[][];
-    // todo: scheduled: IScheduleCourse[][];
```
bugsalexander: I think I mentioned to you at the last last meeting (not 10/20, but the one before) that this was the old Schedule data-definition, or at least I hope I did. In a more recent pull request, I moved the data definitions around so that the ones we're using are at the top, now merged into master. Message me on slack if you have any questions! I'm happy to help.

backend/src/types.ts
```diff
@@ -267,19 +267,30 @@ export interface INEUPrereqCourse {
...
+ * @param title The name of the course, ex "Logic and Computation"
+ * @param dndId A unique id for the course for drag and drop purposes, ex "class-1"
  */
 export interface IScheduleCourse {
-    classId: number;
-    subject: string;
+  classId: number;
+  subject: string;
+  title: string;
+  dndId: string;
```
bugsalexander: I'm not sure what this is, but it may exist on the more detailed interface for a course that we currently use as part of a schedule, `ScheduleCourse`.

backend/src/types.ts
```diff
@@ -288,3 +299,15 @@ export interface IScheduleCourse {
...
  */
 export type UserChoice = ICourseRange | IOrCourse;
+
+/**
+ * A semester.
+ * @param id The id of the semester
+ * @param title The name of the semester: 'Fall', 'Spring'
+ * @param classIds A list of the class dndId's in this semester
+ */
+export interface ISemester {
```
bugsalexander: This also already exists on the master branch as `ScheduleTerm`.

frontend/src/models/types.ts
```diff
@@ -189,7 +189,7 @@ export interface Schedule {
   yearMap: {
```
yasz24: On moving this file to the frontend--I believe this is a limitation of `create react app` since it configures the `webpack` to look for file only in the `frontend/src`. I think there's two options for us here. The first would be to `eject` which will then expose the `webpack.config.js` to us and we could then specify the root from where it should start looking for files. The second option would be to try something like this: `https://medium.com/@ryoldash/customize-webpack-config-of-react-app-created-with-create-react-app-7a78c7849edc ` using the `react-app-rewired` module to override `webpack.config.js`.

I'm not sure if this would work but we could also try restructuring the project as:
├───src/
│   ├───frontend
│   └───backend

bugsalexander:
Also if you want a sample schedule to use for the frontend, there's a couple [here](./backend/test/mock_schedules)

PR #41, opened 2019-10-18
Adding the RANGESection type to the sectionMap

yasz24:


backend/src/types.ts
```diff
@@ -339,6 +339,22 @@ export interface ORSection {
...
+ * A section that has a credit requirement, that can be fulfilled by taking courses in any of the range requirements.
+ * @param type the type of this requirement
+ * @param requirements the possible choices for earning credits
+ * @param numCreditsMin the minimum number of credits needed to satisfy this major
+ * @param numCreditsMax the maximum number of credits needed to satisfy this major
+ * @param name the name of this section
+ */
+export interface RANGESection {
+  type: "RANGE";
+  requirements: Requirement[];
```
bugsalexander: Since we know that this is a RANGEsection, why not make this just a single `ICourseRange`?

backend/src/types.ts
```diff
@@ -304,7 +304,7 @@ export type UserChoice = ICourseRange | IOrCourse;
 export interface Major {
   name: string;
   sections: string[];
-  sectionMap: { [key: string]: ANDSection | ORSection };
+  sectionMap: { [key: string]: ANDSection | ORSection | RANGESection };
```
bugsalexander: At the meeting, you mentioned renaming this to something, will that be updated here?

PR #43, opened 2019-11-09
Add a class to the schedule

NilsBacke:
Styling isn't perfect so it doesn't look great but all the functionality is there :)


bugsalexander:
Picking up the most-recently-added course of the most-recently-added-to semester (abbreviated MRACOTMRATS, or MRAC), causes all other added courses to disappear.
* Upon dropping the MRAC into a term, the others then re-appear. 
* Sometimes grabbing an added course actually picks up the MRAC instead.
* The positions of the other added courses relative to the MRAC appear to stay the same, when moving a non-added course (CS3500) causes the MRAC to shift up/down in the column.

NilsBacke:
> Picking up the most-recently-added course of the most-recently-added-to semester (abbreviated MRACOTMRATS, or MRAC), causes all other added courses to disappear.
> 
> * Upon dropping the MRAC into a term, the others then re-appear.
> * Sometimes grabbing an added course actually picks up the MRAC instead.
> * The positions of the other added courses relative to the MRAC appear to stay the same, when moving a non-added course (CS3500) causes the MRAC to shift up/down in the column.

Yup! An interesting bug. Alas, I fixed it in the commit I just pushed

PR #44, opened 2019-11-09
Port over pre-req validation code to the frontend.

yasz24:


yasz24:
It seems like the `generate-warnings.ts` file isn't being delivered to the browser for some reason (you can click the deployment url, and look at developer tools to find the src files that were served) So this isn't ready to merged yet. Will work on fixing this.

yasz24:
So looks like the file only gets served when the react app actually imports AND uses functions from it. So it should be okay to merge this because Nils will eventually import and use code from the `generate-warnings.ts` file

PR #45, opened 2019-11-09
Add "Completed Courses" modal

NilsBacke:


frontend/src/components/DropDownModal.tsx
```diff
@@ -0,0 +1,62 @@
...
+  render() {
+    return (
+      <Wrapper>
+        <TopInfo onClick={this.handleClick.bind(this)}>
+          <p>4 courses</p>
+          <p>16 credits</p>
+          <p>Freshman Standing</p>
+          <ExpandMoreOutlinedIcon></ExpandMoreOutlinedIcon>
+        </TopInfo>
+        {this.state.expanded && <BottomInfo></BottomInfo>}
```
yasz24: I think we ask users what courses they've already taken during the on-boarding process right? Do we have a way to store that info currently? Are we able to populate that information in the modal? (and also populate their schedule with those courses.)

NilsBacke: Not currently no, but the goal of this PR is to just get the UI in place, and we can add the functionality later once we have the onboarding process done.

PR #49, opened 2019-11-10
Loader

bugsalexander:
This code works, but should be cleaned up before being merged.
* get rid of string interpolation
* use `apollo-link` to make requests, rather than `https`


backend/src/prereq_loader.ts
```diff
@@ -0,0 +1,250 @@
...
+/**
+ * Asynchronously adds prereqs to a Schedule.
+ * Does not do mutation.
+ * @param schedules the schedule to add prereqs to
+ */
+export async function addPrereqsToSchedules(
+  schedules: Schedule[]
+): Promise<Schedule[]> {
+  // set the total to zero
+  totalNumRequests = 0;
```
yasz24: Seems like this was only so you could print the number of requests you ended up making, in which case you should probably just remove this

backend/src/prereq_loader.ts
```diff
@@ -0,0 +1,250 @@
...
+  for (const year of schedule.years) {
+    newYearMap[year] = await prereqifyScheduleYear(
+      schedule.yearMap[year],
+      loader
+    );
+  }
+
+  // reconstructs a year
+  return {
+    years: schedule.years.slice(0),
```
yasz24: isn't this the same as `years: schedule.years` ?

backend/src/prereq_loader.ts
```diff
@@ -0,0 +1,250 @@
...
+ * Queries SearchNEU using GraphQL to look up the prereqs for each of the provided courses.
+ * @param courses the courses to lookup prereqs for
+ */
+async function queryCoursePrereqData(
+  courses: SimpleCourse[]
+): Promise<PrereqQueryResult[]> {
+  // for each one of the courses, map to a string.
+  const courseSchema: string[] = courses.map((course: SimpleCourse) => {
+    return `class(classId: ${course.classId}, subject: "${course.subject}") { 
+      occurrence(termId: ${202010}) {
```
yasz24: why is the termId set to `202010` here specifically?

backend/src/prereq_loader.ts
```diff
@@ -0,0 +1,250 @@
...
+  // make the request.
+  let queryResult = await request({
+    uri: "https://searchneu.com/graphql",
+    method: "POST",
+    body: JSON.stringify({ query: querySchema }),
+    headers: {
+      "Content-type": "application/json",
+    },
+  });
+  totalNumRequests += 1;
```
yasz24: Same as above. Out of curiosity, how many requests does it end up making? It it called once for the length of the `SimpleCourse[]` or fewer times?

backend/src/prereq_loader.ts
```diff
@@ -0,0 +1,250 @@
...
+  };
+}
+
+/**
+ * Asynchronously adds prereqs to a provided course.
+ * @param course the course to add prereqs to
+ * @param loader the loader
+ */
+async function prereqifyScheduleCourse(
+  course: ScheduleCourse,
```
yasz24: I don't know if you already tried this but maybe you could `console.log(currentTime)` in here. This would tell us what the time intervals are between the calls to `prereqifyScheduleCourse` for each `ScheduleCourse` in our array. It could give us some insights on the number of queries in a single `DataLoader` batch.

yasz24:
Just had some minor stuff and a few comments. But this looks awesome otherwise.

backend/src/prereq_loader.ts
bugsalexander: It makes around 20 per unique plan! For BSCS, made 20 requests for the first plan, and then zero for the second and third.

backend/src/prereq_loader.ts
bugsalexander: Yes, almost. I had to do this because `schedule.years` for some reason didn't always have the actual `Array` prototype. Doing this converts from the fake proto to the real one.

That being said, I changed it back and nothing broke, so idk!

backend/src/prereq_loader.ts
bugsalexander: Given a plan with courses that don't have prereqs, I decided that we should look at the most recent year of the catalog to grab their prereqs.

However, I'm going to change this to be year-specific (pass in a year along with the plan to grab prereqs for that plan, from that year of the catalog).

backend/src/prereq_loader.ts
bugsalexander: Seems to be a couple of milliseconds. 

The "want a course" means `prereqifyScheduleCourse` was called, while "request a course" means the load function `queryCoursePrereqData` was called.
Screen Shot 2019-11-11 at 3 40 09 PM
Screen Shot 2019-11-11 at 3 40 15 PM
Screen Shot 2019-11-11 at 3 40 23 PM

PR #52, opened 2019-11-14
height of all semesters grow as one unit

NilsBacke:


NilsBacke:
> Wonderful! Was it really a one-line fix/adding semicolons?

Yup! Just needed to add 'height: 100%' in the right place

PR #62, opened 2019-11-28
Course names

bugsalexander:
Added course names!!! Now they display with all their glory :)))

Full list of additions:
* improved the elective names scraped by the plan_scraper
* loader now uses minCreds, maxCreds, name from SearchNEU 
* updated frontend copies of the schedules 
* added more tests for prereq_loader
* migrated tests that were basically snapshot tests, to actual snapshots


bugsalexander:
Thanks Nils!!

PR #63, opened 2019-11-29
Major Requirements Sidebar

NilsBacke:


frontend/src/utils/generate-warnings.ts
```diff
@@ -184,6 +184,9 @@ function produceUnsatifiedRequirement(
   taken: Map<string, HashableCourse>,
   coursesUsed: Set<string>
 ): IRequirementGroupWarning | undefined {
+  if (!requirementGroup) {
+    return undefined;
+  }
```
bugsalexander: Why did you make this change? I just deployed after removing it, and it worked fine.

bugsalexander:
Also, would it be possible to make the checkmarks green and the X's red?

frontend/src/utils/generate-warnings.ts
NilsBacke: You're right! I think I came upon an error before where it was undefined, but I think I was doing something else wrong.

bugsalexander:
The biochem major had a bug – whoops! Yash is fixing it now. 

In other news, courses that are added to the schedule via the + buttons aren't taken into account when producing warnings. Any idea on why?

bugsalexander:
Just merged in master to get the hotfix that @yasz24 pushed for the Biochem major. Now selecting biochem no longer crashes the site!

PR #70, opened 2019-12-03
missing prereqs warning bugfix

bugsalexander:
Fixed the last bug for frontend warnings not working on added courses.


bugsalexander:
As a sidenote, it's pretty dumb that in the graphql query, the credit props are called `maxCredits` and `minCredits`, and our data definition calls them `numCreditsMax` and `numCreditsMin`. I opened another trello card for that.

PR #72, opened 2019-12-05
Onboarding questions

NilsBacke:


ArunJ98:
Some thoughts - It would be nice to have some validation for the inputs, for example, you can click through to next with an empty name. Other fields like graduation year could also be validated. Also, should the expected graduation field have a dropdown similar to selecting Academic year?

NilsBacke:
> Some thoughts - It would be nice to have some validation for the inputs, for example, you can click through to next with an empty name. Other fields like graduation year could also be validated. Also, should the expected graduation field have a dropdown similar to selecting Academic year?

Great points! Validation is something I overlooked for sure. I was thinking expected graduation would just be a textfield because otherwise we would have to update it once a year to add the new freshman class graduation year. I'll put up a new commit with some changes shortly.

frontend/src/Onboarding/MajorScreen.tsx
```diff
@@ -0,0 +1,72 @@
...
+        options={majors.map(maj => maj.name)}
+        renderInput={params => (
+          <TextField
+            {...params}
+            variant="outlined"
+            label="Select A Major"
+            fullWidth
+          />
+        )}
+        value={!!this.state.major ? this.state.major.name + " " : ""}
```
yasz24: For some reason this doesn't seem to be working right. I don't see the major.name populated when it's selected from the drop-down

yasz24: Also why the `!!` ?

NilsBacke: This is really weird. It works fine for me when I test it locally using `yarn start`, but it doesn't work on zeit.

frontend/src/home/Home.tsx
```diff
@@ -364,6 +362,7 @@ class HomeComponent extends React.Component<Props, HomeState> {
   }
 
   render() {
+    console.log("planStr " + this.state.planStr);
```
yasz24: maybe get rid of this console.log? looks great otherwise

PR #73, opened 2019-12-07
Add course-specific warnings

bugsalexander:
This PR adds course-specific warnings both as a type, and implements them for prereq and coreq checking. 

Also introduces a warning container (object) that can hold both normal and course specific warnings.


frontend/src/home/Home.tsx
```diff
@@ -186,16 +186,17 @@ class HomeComponent extends React.Component<HomeProps, HomeState> {
   }
 
   updateWarnings(newState: HomeState) {
-    const warnings = produceWarnings(newState.schedule);
+    const container = produceWarnings(newState.schedule);
+    const normal = container.normalWarnings;
```
bugsalexander: @NilsBacke, this `container` variable here is an object, with 2 fields: normalWarnings, and courseWarnings. The prior are normal warnings, and I changed this code to handle them like normal (hah!). The `container.courseWarnings` field contains a list of `CourseWarning`, each of which is a normal warning with the added properties of `subject` and `classId`. 

Hopefully this is what you need!

bugsalexander:
This branch should not be merged until @NilsBacke has added highlighting for course-specific warnings.

frontend/src/components/ClassBlocks/ClassBlock.tsx
```diff
@@ -16,20 +16,23 @@ const Block = styled(Card)<any>`
...
+  background-color: ${props =>
+    props.warning ? "rgba(216, 86, 86, 0.9)" : "rgba(173, 198, 255, 0.3)"};
   padding-left: 8px;
   flex: 1;
 `;
 
 interface ClassBlockProps {
   class: DNDScheduleCourse;
   index: number;
+  warning: boolean;
```
bugsalexander: Could you change the type of warning to `string` instead, and store the message here, if it has one?

frontend/src/components/SemesterBlock.tsx
```diff
@@ -58,12 +63,24 @@ export class SemesterBlock extends React.Component<
...
-            <ClassBlock key={index} class={scheduleCourse} index={index} />
+            <ClassBlock
+              key={index}
+              class={scheduleCourse}
+              index={index}
+              warning={
+                !!courseWarnings.find(
+                  w =>
+                    w.subject + w.classId ===
+                    scheduleCourse.subject + scheduleCourse.classId
```
bugsalexander: might also need to change this bit to either return the message, or store the entire warning

PR #78, opened 2019-12-15
Redux

NilsBacke:
Revamping our state management system. This should make it a lot easier to add new features in the future.

Added local storage as well (for the whole AppState)


frontend/package.json
```diff
@@ -7,26 +7,36 @@
     "@material-ui/icons": "^4.5.1",
     "@material-ui/lab": "^4.0.0-alpha.31",
     "@material-ui/styles": "^4.7.1",
+    "@react-native-community/async-storage": "^1.7.1",
```
bugsalexander: This isn't used anywhere, but feel free to ignore if it's needed for something else not specifically imported or something.

frontend/src/App.tsx
```diff
@@ -7,21 +7,33 @@ import { AcademicYearScreen } from "./Onboarding/AcademicYearScreen";
...
-          <Route path="/minors" component={MinorsScreen} />
-          <Route path="/" component={Onboarding} />
-        </Switch>
-      </Router>
-    );
-  }
-}
+export const App = ({
+  store,
+  persistor,
```
bugsalexander: Should there be type annotations here?

frontend/src/components/Sidebar/RequirementSection.tsx
```diff
@@ -94,10 +94,12 @@ export class RequirementSection extends React.Component<
   }
 
   parseRequirements(reqs: Requirement[]) {
-    return reqs.map(r => this.renderRequirement(r));
+    return reqs.map((r: Requirement, index: number) => (
+      <div key={index + 10000}>{this.renderRequirement(r, index)}</div>
```
bugsalexander: Why plus 10000?

frontend/src/components/Sidebar/RequirementSection.tsx
```diff
@@ -121,17 +123,21 @@ export class RequirementSection extends React.Component<
...
-          .map(c => this.renderRequirement(c))}
+          .map((c: Requirement, index: number) =>
+            this.renderRequirement(c, index)
+          )}
         {req.courses
           .filter(c => c.type === "OR")
-          .map(c => this.renderRequirement(c))}
+          .map((c: Requirement, index: number) =>
+            this.renderRequirement(c, index)
+          )}
```
bugsalexander: This should probably go in a separate PR, but could you change the requirements to be displayed using prefix + parens? Ex. (OR (AND CS2500 CS2501) EECE2160). The parens really clear stuff up, rather than having to look carefully at indentation, for example with the science requirement of CS.

frontend/src/components/Sidebar/Sidebar.tsx
```diff
@@ -33,7 +36,10 @@ export const Sidebar: React.FC<Props> = ({ schedule, major }) => {
     );
   }
 
-  const warnings = produceRequirementGroupWarning(schedule, major);
+  const warnings = produceRequirementGroupWarning(
+    JSON.parse(JSON.stringify(schedule)),
+    major
+  ); // deep copy of schedule
```
bugsalexander: why deep copy of schedule? If produceRequirementGroupWarning somehow mutates, PLEASE let me know (I need to change that). If it's because the produced warnings contain references to things in the schedule, also please let me know (I should change that too).

frontend/src/utils/schedule-helpers.ts
```diff
@@ -107,10 +107,10 @@ export const convertToDNDSchedule = async (
   return [newSchedule, counter];
 };
 
-export const convertToDNDCourses = async (
+export const convertToDNDCourses = (
   courses: ScheduleCourse[],
   counter: number
-): Promise<[DNDScheduleCourse[], number]> => {
+): [DNDScheduleCourse[], number] => {
   var list: DNDScheduleCourse[] = [];
```
bugsalexander: Can you replace `var` with `const`?

tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
-    "target": "es5",
-    "lib": ["dom", "dom.iterable", "esnext"],
+    "target": "es6",
+    "lib": ["dom", "dom.iterable", "esnext", "es6"],
```
bugsalexander: I'm sure this is necessary, but could you explain? (this is mostly for my understanding).

frontend/package.json
NilsBacke: You're right!

frontend/src/App.tsx
NilsBacke: There are. On lines 19-20 is the type annotation. Kind of a pain but this is the syntax that works with typescript. First define the object without types, then define the object with types.

frontend/src/components/Sidebar/RequirementSection.tsx
NilsBacke: This is old, it was from when I was messing around with the keys. Good catch!!

frontend/src/components/Sidebar/RequirementSection.tsx
NilsBacke: Made a card for this in Trello

frontend/src/components/Sidebar/Sidebar.tsx
NilsBacke: I was running in to trouble with this because since `schedule` is from redux, it cannot be directly mutated. If I remember correctly it was caused by a `.sort` call.

frontend/src/utils/schedule-helpers.ts
NilsBacke: Yup! Thanks

tsconfig.json
NilsBacke: I remember I had a reason for doing this, but I just tried it with es5 and everything seems fine. Must've been something I tried that didn't work for something else.

frontend/src/App.tsx
bugsalexander: OH I see, confused that with the return type. Thanks!

frontend/src/components/Sidebar/Sidebar.tsx
bugsalexander: Thanks, I'll look into it. It shouldn't be mutating, so that sounds like an error.

tsconfig.json
bugsalexander: I'd expect you did too, so maybe it'll come back to bite us. Leaving this conversation unresolved just in case we run into issues later on.

PR #79, opened 2019-12-15
Add visible warning to a semester with a warning

NilsBacke:
Note: This PR contains all the changes from the Redux PR. 

This should be merged after the redux PR.

### DIFF
--- a/frontend/src/components/SemesterBlock.tsx
+++ b/frontend/src/components/SemesterBlock.tsx
@@ -48,7 +48,7 @@ const NoClassBlock = styled.div`
 
 interface ReduxStoreSemesterBlockProps {
   courseWarnings: CourseWarning[];
-  warning?: IWarning;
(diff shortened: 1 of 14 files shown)


frontend/src/components/SemesterBlock.tsx
```diff
@@ -136,29 +162,27 @@ class SemesterBlockComponent extends React.Component<
...
 
 const mapStateToProps = (state: AppState, ownProps: SemesterBlockProps) => ({
+  warning:
+    getWarningsFromState(state).filter(
+      w => w.termId === ownProps.semester.termId
+    ).length > 0
+      ? getWarningsFromState(state).filter(
+          w => w.termId === ownProps.semester.termId
+        )[0]
+      : undefined,
```
bugsalexander: Rather than just grabbing [0] here, should probably just be the entire filtered array.

PR #82, opened 2019-12-16
Fetch majors/plans from searchNeu

yasz24:


yasz24:
@bugsalexander 
The additional requests are because we make `CORS` requests as per here:(https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS) since we want to fetch from the domain `searchneu.com` which is not the same as the domain from where the javascript for our app was served `(graduatenu.com)`. Further we fall under this category from the article: `Invocations of the XMLHttpRequest or Fetch APIs`, because our app is a react app and doesn’t do a full page reload. Further, we need to set the content-type to `application/json` which is not one of the allowed types to be classified as a simple request. 

It seems like the browser will automatically trigger this “pre-flight” `OPTIONS` request as a pre-caution and there is nothing we can do about it. 

As for the scalability concern I profiled the load times, for loading 1 copy vs 128 copies of the`bio-chem` major, (I chose bio-chem because most no. of plans). I was looking at the request completion times for the `xhr` requests in the network tab in chrome dev tools:
Screen Shot 2019-12-22 at 8 15 11 AM

1 copy of biochem:
- majorRequest, load times (10 loads): 0.36, 1.07, 0.41, 0.41, 0.41, 1.23, 1.13, 0.41, 0.47, 0.48, 0.34
- majorRequest, avg. load time:  0.67s
- plansRequest, load times (10 loads): 1.09, 1.08, 0.39, 0.41, 1.21, 0.41, 0.41, 1.13, 0.41, 0.4
- plansRequest, avg. load time: 0.7s
Time the user had to wait on average: max(0.67, 0.7) = `0.7s`

128 copies of biochem:
- majorRequest, load times (10 loads): 2.66, 3.28, 1.64, 3.88, 1.33, 2.89, 1.85, 1.27, 1.33, 2.69
- majorRequest, avg. load time:  2.28s 
- plansRequest, load times (10 loads): 2.56, 2.58, 3.26, 1.48, 2.87, 3.07, 3.74, 3.67, 3.69, 3.69
- plansRequest, avg. load time: 3.06s
Time the user had to wait on average: max(2.28, 3.06) = `3.06s`

Since we are requesting the same major object 128 times, the same SQL query will be repeatedly executed on the search backend. I’m guessing (I hope this is the case) that `Postgres` will cache the result of the first query, speeding up the execution of the ones that follow. If we instead had 128 distinct major objects, it would mean that 128 distinct SQL queries would have to be executed which would technically be slower than the profiled time for 128 copies of the same major(3.06s). However, this is a classic case of the N+1 problem, and can be solved by eager loading with `include` in the resolvers  for the graphql (https://stackoverflow.com/questions/53905806/understanding-include-when-resolving-source-association-from-target-in-sequeli), which should keep the avg. wait time around the 3.06s, even for 128 distinct majors. Unfortunately, since we do not currently support 128 distinct majors, we cannot actually profile this right now. 

I think what we can conclude from this experiment is that the overhead per additional major requested isn’t all that much. So, from a load time stand point, we’re probably okay scaling to 100+ majors. 

With storage, the size of the JSON response(for 128 objects) is only `1.8kb` and `4.2kb` for the majors and plans respectively. I don’t think the overhead to represent that JSON as major and schedule object types and storing them in memory would be that much more.  

What I’m not sure of is which one is the better user experience. Front-load fetching the plans/majors and have users wait `3s` once, but then have the ability to seamlessly switch between majors without any additional wait? Or have them not wait at all initially, but wait `0.7s` each time they want to try a different major, the time for which would add up. Another con of the latter approach is that we make a network request each time, which increases the server-load for no. of requests to be processed. Some users probably won’t switch between too many majors, but others might want to try several (if they are undecided freshman for example.)

In my head I can make an argument for both approaches. But, let me know what you think.

yasz24:
Combined the requests for majors and plans to only make a single request now.  Further, the `fetch` for `majors` and `plans` is now triggered as soon as the app is served and continues in the background as the user navigates through the first few on-boarding screens. This process should take the user at least the  `3.06s` which is the worst case for fetching 100+ `majors` and `plans`. A spinner will only be displayed if the user gets to the `MajorScreen` before the `fetch` was completed. This design doesn't block the rest of the UI from loading.

yasz24:
[Use the Search end-point to get majors/plans](https://trello.com/c/FVpaIM0L/27-use-the-search-end-point-to-get-majors-plans)

frontend/src/state/reducers/scheduleReducer.ts
```diff
@@ -120,7 +120,7 @@ export const scheduleReducer = (
         return draft;
       }
       case getType(setScheduleAction): {
-        const { schedule } = action.payload;
+        const schedule = JSON.parse(JSON.stringify(action.payload.schedule)); /// deep copy of schedule, because schedule is modified
```
bugsalexander: Why didn't we need a deep copy before, but do now?

frontend/src/utils/fetchMajorsAndPlans.ts
```diff
@@ -0,0 +1,84 @@
...
+import { Schedule } from "../models/types";
+
+const majorSchema: string[] = majorIds.map((majorId: string) => {
+  return `major(majorId: "${majorId}") {
+        latestOccurrence {
+            requirements
+            plansOfStudy
+        }
+    }`;
+});
```
bugsalexander: This should be done using a GraphQL client like Relay or Apollo, should be done in a separate issue.

bugsalexander:
Also – do you think you could go back and leave comments annotating what each bit of the code does?

frontend/src/state/reducers/scheduleReducer.ts
yasz24: I mentioned this to you at the meeting, but it has something to do with React props being immutable, which caused an error.

frontend/src/utils/fetchMajorsAndPlans.ts
yasz24: I will make a card for this on our trello board

PR #89, opened 2020-01-25
Implementing an error message for adding duplicate classes

imanmoreira:


yasz24:
[Add a warning for duplicate classes in a schedule](https://trello.com/c/0i6xayHr/43-add-a-warning-for-duplicate-classes-in-a-schedule)

frontend/src/components/AddClassModal.tsx
```diff
@@ -1,9 +1,21 @@
...
+import {
+  ScheduleCourse,
+  DNDSchedule,
+  DNDScheduleTerm,
+  DNDScheduleCourse,
+} from "../models/types";
 import { XButton } from "./common";
 import { fetchCourse } from "../api";
 import { Modal, CircularProgress } from "@material-ui/core";
+////
```
yasz24: You should take out these `////` comments

frontend/src/components/AddClassModal.tsx
```diff
@@ -1,9 +1,21 @@
...
+import { AppState } from "../state/reducers/state";
+import { getScheduleFromState } from "../state";
+import { connect } from "react-redux";
+import { withRouter, RouteComponentProps } from "react-router-dom";
+import { withToast } from "../home/toastHook";
+////
```
yasz24: same as above

frontend/src/components/AddClassModal.tsx
```diff
@@ -138,6 +158,33 @@ export class AddClassModal extends React.Component<
     }
   }
 
+  isCourseInSchedule(courseToAdd: ScheduleCourse, schedule: DNDSchedule) {
```
yasz24: I would add java doc style comments explaining what each of the functions that you wrote do

frontend/src/utils/generate-warnings.ts
```diff
@@ -676,13 +689,64 @@ function produceSpecificCourseWarnings(
   warnings = warnings.concat(
     checkPrerequisites(term.classes, tracker, term.termId)
   );
+  ///// ADD CODE HERE TO CHECK IF THERE ARE DUPLICATES
```
yasz24: you should take out this comment

frontend/src/utils/generate-warnings.ts
```diff
@@ -676,13 +689,64 @@ function produceSpecificCourseWarnings(
...
+  warnings = warnings.concat(
+    checkDuplicates(term.classes, tracker, term.termId)
+  );
+  ////////////// END ///////////
```
yasz24: take out this comment

frontend/src/utils/generate-warnings.ts
```diff
@@ -676,13 +689,64 @@ function produceSpecificCourseWarnings(
...
+function checkDuplicates(
+  toCheck: ScheduleCourse[],
+  tracker: CourseTakenTracker,
+  termId: number
+): CourseWarning[] {
+  const warnings: CourseWarning[] = [];
+  for (let i = 0; i < toCheck.length; i++) {
+    const course = toCheck[i];
+    if (
+      courseCode(course) != "XXXX9999" &&
```
yasz24: actually, now that i think of it you should probably define an array of strings at the top of this file, that contains the names of courses that are placeholders

frontend/src/utils/generate-warnings.ts
```diff
@@ -676,13 +689,64 @@ function produceSpecificCourseWarnings(
...
+        subject: course.subject,
+        classId: course.classId,
+        message: `${courseCode(course)}: is in your schedule multiple times`,
+        termId: termId,
+      });
+      warnings.push({
+        subject: course.subject,
+        classId: course.classId,
+        message: `${courseCode(course)}: is in your schedule multiple times`,
+        termId: tracker.getTermId(courseCode(course)),
```
yasz24: I was curious as to why you add two warnings for the same course here. Wouldn't a single warning be sufficient.

yarn.lock
```diff
@@ -9555,14 +9560,7 @@ redux-persist@*, redux-persist@^6.0.0:
...
 
-redux-undo@^0.6.1:
-  version "0.6.1"
-  resolved "https://registry.yarnpkg.com/redux-undo/-/redux-undo-0.6.1.tgz#ff407749b8f468bead636e413819e92c0982eeca"
-  integrity sha1-/0B3Sbj0aL6tY25BOBnpLAmC7so=
-  dependencies:
-    redux "^3.0.2"
-
-redux@^3.0.2, redux@^3.6.0:
+redux@^3.6.0:
```
yasz24: I believe this change is not required.

package.json
```diff
@@ -33,5 +33,9 @@
     "hooks": {
       "pre-commit": "pretty-quick --staged"
     }
+  },
```
yasz24: I believe prettier and react are already dependencies in `frontend/package.json` so you shouldn't need this here

frontend/src/utils/generate-warnings.ts
imanmoreira: So the reason for this is because the "first" of the duplicates does not think it is a duplicate, because it only sees course in the schedule before this class, so when the second class notices the first as a duplicate, it needs to mark itself and the other one as a duplicate. I just based it on the structure of the current code because right now the tracker only knows the course before the term you're on.

PR #91, opened 2020-02-16
Implemented sidebar add buttons

liustanley:


frontend/src/components/Sidebar/SidebarAddClassModal.tsx
```diff
@@ -0,0 +1,257 @@
...
+
+  prepareToClose() {
+    this.setState({
+      queuedCourses: [],
+    });
+    this.props.handleClose();
+  }
+
+  mapScheduleToSemesters() {
+    let semesterList = [];
```
yasz24: These variables should all be type annotated. Type annotations prevent many runtime bugs, by catching them at compile time and it also makes the code more comprehensible.

frontend/src/components/Sidebar/SidebarAddClassModal.tsx
```diff
@@ -0,0 +1,257 @@
...
+          </form>
+          <SubmitButton onClick={this.handleSubmit.bind(this)}>
+            Add Class
+          </SubmitButton>
+        </InnerSection>
+      </Modal>
+    );
+  }
+}
+
```
yasz24: you should add brief comments about what these objects do in the context of connect. as an example you can check. `https://github.com/sandboxnu/graduatenu/blob/fetch-majors-plans/frontend/src/Onboarding/MajorScreen.tsx#L166`

yasz24:
[Add "add" button to the right sidebar for each class](https://trello.com/c/0InO3JBW/39-add-add-button-to-the-right-sidebar-for-each-class)

PR #99, opened 2020-03-04
Users Endpoints

yasz24:


yasz24:
[Users endpoint](https://trello.com/c/tGkRHaF7/63-users-endpoint)

api/app/controllers/application_controller.rb
```diff
@@ -1,2 +1,51 @@
...
+            end
+          end
+        end
+    end
+
+    def authenticate_user!(options = {})
+        head :unauthorized unless signed_in?
+    end
+
+    def current_user
```
William-Cunningham, line 44: Just thought it might be nice to add a comment here since this line is so critical for the controller actions.

api/app/controllers/users_controller.rb
```diff
@@ -0,0 +1,20 @@
+class UsersController < ApplicationController
+    before_action :authenticate_user!
+    def show
```
William-Cunningham, line 3: I'm unclear on exactly why the show action has no code but this may just be from not knowing rails all that well. But even so, it might be nice to add a comment explaining why just for readability?

PR #103, opened 2020-03-22
Catalog scraper tests

imanmoreira:
Several changes to the catalog scraper, such as modifying the way it parses or courses, changing the way Ranges are parsed to allow rows that have no course links on it and adding in an end range, allowing for ranges specified by 'any ____ course' type of rows, making some small alterations to subheader parsing and some other smaller bugs. 
Also added several new majors.

### DIFF
--- a/frontend/src/models/types.ts
+++ b/frontend/src/models/types.ts
@@ -47,6 +47,8 @@ export type Requirement =
  */
 export interface IOrCourse {
   type: "OR";
+  //numCreditsMin: number;
+  //numCreditsMax: number;
   courses: Requirement[];
 }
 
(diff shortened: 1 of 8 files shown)


yasz24:
[support more majors on the backend](https://trello.com/c/IQZ14ZgF/18-support-more-majors-on-the-backend)

backend/src/catalog-scraper/reqGroup_scraper.ts
```diff
@@ -135,31 +174,21 @@ function processAndSection(
     //can parse the rows as indivdual requirements
     let containsOrRow: boolean = false;
```
yasz24, line 175: you can get rid of this flag, so it doesn't confuse others

PR #110, opened 2020-04-05
Onboarding screen 1

imanmoreira:
created the redesign of the first onboarding information screen


frontend/src/components/common/NextButton.tsx
```diff
@@ -1,12 +1,21 @@
...
 import { Button } from "@material-ui/core";
+import { red, green, blue, grey } from "@material-ui/core/colors";
 
 export const NextButton: React.FC = () => {
   return (
     <Button
       variant="contained"
       color="secondary"
-      style={{ marginTop: 12, marginBottom: 12 }}
+      style={{
```
liustanley, line 10: Using inline styles to this extent is generally considered bad practice, it would be better to convert this Button into a styled-component.

frontend/package.json
```diff
@@ -57,7 +57,7 @@
   },
   "scripts": {
     "start": "react-scripts start",
-    "build": "react-scripts build",
+    "build": "CI=false react-scripts build",
```
liustanley, line 60: I'm not sure if this change was intended, and if it wasn't then it might be worth a second look to make sure it doesn't break any tests or anything.

yasz24:
["Let's get to Know you" - Onboarding](https://trello.com/c/WoB4Smva/69-lets-get-to-know-you-onboarding)

frontend/package.json
imanmoreira: This was intentional -- It wasn't running the deployments correctly without that

PR #112, opened 2020-04-08
Run postgres server in docker for local dev environment

yasz24:


yasz24:
[Run postgres server in a docker container for local dev-environment](https://trello.com/c/RvP2trH5/75-run-postgres-server-in-a-docker-container-for-local-dev-environment)

api/config/database.yml
```diff
@@ -20,8 +20,10 @@ default: &default
   # For details on connection pooling, see Rails configuration guide
   # https://guides.rubyonrails.org/configuring.html#database-pooling
   pool: <%= ENV.fetch("RAILS_MAX_THREADS") { 5 } %>
-  username: api
-  password: <%= ENV['API_DATABASE_PASSWORD'] %>
+  username: postgres
```
ArunJ98, line 23: should these be ENV variables?

yasz24: Yea i thought about that and we could, but then developers have to go through the additional step of setting up env variables in their `bash_profile` or something, and make sure those are set up correctly, where as right now they can just plug and play `yarn start` after they pull from master. Since this is just for the development i thought we shouldn't care that much. Definitely something to do for production though

PR #115, opened 2020-04-19
Make new majors live

yasz24:


yasz24:
[Make new majors and plans live on the app](https://trello.com/c/hyzyOWRP/88-make-new-majors-and-plans-live-on-the-app)

PR #117, opened 2020-04-30
Containerize rails app

yasz24:


yasz24:
[Containerize Rails app/frontend for deployment](https://trello.com/c/QjyX8WhW/76-containerize-rails-app-frontend-for-deployment)

PR #118, opened 2020-04-30
Populate schedule from excel plan of study

ArunJ98:
Populates the schedule by parsing the excel plan of study from advisors.


frontend/src/components/ExcelUpload.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+const Container = styled.div`
+    position: relative;
+    justify-content: right;
+    padding: 0px;
+    margin: 0px;
+`;
+
+
+interface props {
```
NilsBacke, line 18: Me being picky but interface names are generally capitalized

frontend/src/components/ExcelUpload.tsx
```diff
@@ -0,0 +1,57 @@
...
+    setSchedule: ((schedule: Schedule) => void)
+}
+
+export class ExcelUpload extends React.Component<props> {
+    constructor(props: any) {
+        super(props);
+        this.handleChange = this.handleChange.bind(this);
+    }
+
+    async handleChange(e : any) {
```
NilsBacke, line 28: Again this is super minor but I think a better name for this function would be `onUpload`

frontend/src/home/Home.tsx
```diff
@@ -41,6 +42,7 @@ import {
 import { setMajorAction } from "../state/actions/userActions";
 import { getMajors, getPlans } from "../state";
 import { EditPlanPopper } from "./EditPlanPopper";
+import { ExcelUpload as ExcelUpload } from "../components/ExcelUpload";
```
NilsBacke, line 45: Do we still need the `as`?

frontend/src/utils/excelParser.ts
```diff
@@ -0,0 +1,250 @@
...
+    SeasonEnum,
+    ScheduleYear,
+    ScheduleTerm,
+    StatusEnum
+} from "../models/types";
+import { convertSeasonToTermId } from './schedule-helpers';
+
+const BASE_YEAR: number = 1000;
+
+export function ExcelToSchedule(file: File, callback: ((schedule: Schedule) => any)) {
```
NilsBacke, line 14: I thought this was a class name because of the capitalization, I think `excelToSchedule` would be more clear

frontend/src/components/ExcelUpload.tsx
```diff
@@ -0,0 +1,57 @@
...
+        if (e != null && e.target != null) {
+            console.log(e.target.files);
+            const file = e.target.files[0]
+            ExcelToSchedule(file, this.props.setSchedule)
```
NilsBacke, line 32: Is it possible to make this an async function and have it return a schedule, instead of having to pass in a callback?

frontend/src/home/Home.tsx
```diff
@@ -266,6 +276,12 @@ class HomeComponent extends React.Component<Props> {
     ));
   }
 
+  setSchedule(schedule: Schedule) {
+    console.log(schedule);
+    const [dndschedule, counter] = convertToDNDSchedule(schedule, 0);
```
yasz24, line 281: We would actually need to `prereqify` the schedule before we set it here. Without this we won't be able to generate the pre-req/co-req warnings for the course objects in the schedule that came from the excel sheets.

To do this you could make use of the utility function `prereqifySchedule` in `scrapers/src/prereq_loader.ts`. You need to provide it with a `DataLoader` object. An example usage of this function is in `addPrereqsToSchedules` in the same file.

frontend/src/components/ExcelUpload.tsx
ArunJ98: Talked on slack and decided to keep the callback

frontend/src/home/Home.tsx
ArunJ98: Added Prerequefying schedule

PR #121, opened 2020-05-08
Refactor types to a common library

ArunJ98:
Currently, only known issue is that I was unable to export enum's probably so they are duplicated.


graduate-common/src/tests/prereq_loader.test.js
```diff
@@ -1,6 +1,6 @@
-const prereq_loader = require("../src/prereq_loader");
+const prereq_loader = require("../prereq_loader");
```
yasz24, line 1: what do you think about moving the `tests` outside of `src` to the top-level of the `graduate-common` package?

ArunJ98: that makes sense to me, will update the PR

PR #130, opened 2020-05-20
Loading and saving plans from API

liustanley:
Adds support for user plans:

- Added major strings to plans API model
- When a user signs up, their first plan will also be created automatically
- If a user is logged in, navigating to /home will fetch and display their plan
- When a user is logged in, pressing "Save Plan" on the home page will update their plan on the backend
    - If a user is logged in, their changes will not be persisted without pressing "Save Plan"
    - If a user is not logged in, their changes will be persisted the same as before


liustanley:
[Loading and saving plans from API](https://trello.com/c/uqeLO3ai/101-loading-and-saving-plans-from-api)

frontend/src/home/Home.tsx
```diff
@@ -277,12 +389,103 @@ class HomeComponent extends React.Component<Props> {
...
+          course_warnings: scheduleData.courseWarnings,
+        }
+      ).then(plan => {
+        this.props.updateActiveSchedule({
+          ...plan.plan,
+          currentClassCounter: plan.plan.courseCounter,
+          isScheduleLoading: false,
+          scheduleError: "",
+        } as ScheduleSlice);
+        alert("Your plan has been updated.");
```
ArunJ98, line 449: Also for a later PR - we should have a better UI than alert here

ArunJ98 (dismissed):
Overall looks great! Left a few comments but most of them are for future PR's. I think the current way we have "save" and "update" is confusing as well but that can be changed later. Maybe we have a major feature branch for this and simply merge future pull requests into that branch before making a final merge to master when everything ready.

imanmoreira (approved):

PR #136, opened 2020-06-07
Render ClassBlock with multiple warnings

liustanley:
Added multi line warning tooltip and moved warning tooltip below the ClassBlock.


liustanley:
[How to render warnings for classes that have more than one](https://trello.com/c/U7vTL1wc/78-how-to-render-warnings-for-classes-that-have-more-than-one)

PR #138, opened 2020-06-14
Plans dropdown

fgridley:
Adds schedule dropdown menu presentational component and provides scaffolding for multiple schedule state management. Future work still required to hook this up with the schedule being displayed.
Screen Shot 2020-06-14 at 13 50 02

Edit: it seems vscode may have done some unintentional autolinting of files not associated with this functionality. I can try to undo it if preferred.


frontend/src/home/SwitchPlanPopper.tsx
```diff
@@ -0,0 +1,150 @@
...
+  activeSchedule: getActiveScheduleFromState(state),
+});
+
+const mapDispatchToProps = (dispatch: Dispatch) => ({
+  setActiveSchedule: (activeSchedule: number) =>
+    dispatch(setActiveScheduleAction(activeSchedule)),
+  addSchedule: (name: string, schedule: PastPresentSchedule) =>
+    dispatch(addScheduleAction(schedule, name)),
+  removeSchedule: (name: string, schedule: PastPresentSchedule) =>
+    dispatch(removeScheduleAction(schedule, name)),
```
imanmoreira, line 139: I don't think you use removeSchedule and addSchedule in here, so it's probably not necessary to keep these fields

frontend/src/models/types.ts
```diff
@@ -41,6 +41,25 @@ export enum SeasonEnum {
...
+export interface PastPresentSchedule {
+  past?: ScheduleSlice;
+  present: ScheduleSlice;
+}
+
+export interface NamedSchedule {
+  name: string;
+  schedule: PastPresentSchedule;
+}
+
```
imanmoreira, line 62: Add some comments about what these data types represent

frontend/src/state/reducers/schedulesReducer.ts
```diff
@@ -0,0 +1,67 @@
...
+  addScheduleAction,
+  removeScheduleAction,
+  setActiveScheduleAction,
+} from "../actions/schedulesActions";
+
+export interface SchedulesState {
+  activeSchedule: number;
+  schedules: NamedSchedule[];
+}
+
```
imanmoreira, line 16: I think schedulesReducer clashes a little too closely with scheduleReducer. Could you change it to something like plansReducer?

frontend/src/state/actions/schedulesActions.ts
```diff
@@ -1,21 +1,4 @@
...
-  })
-)();
-
-export const removeScheduleAction = createAction(
-  "schedules/REMOVE_SCHEDULE",
-  (schedule: PastPresentSchedule, name: string) => ({
-    schedule,
-    name,
-  })
-)();
```
imanmoreira: You removed these from here but I think you still refer to addScheduleAction and removeScheduleAction in the reducer and the SwitchPlanPopper so it's not deploying properly

PR #140, opened 2020-06-19
Classes during coop

William-Cunningham:
pull request for classes during coop

### DIFF
--- a/api/db/schema.rb
+++ b/api/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_05_17_195239) do
+ActiveRecord::Schema.define(version: 2020_06_17_190246) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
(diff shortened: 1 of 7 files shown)


frontend/src/utils/generate-warnings.ts
```diff
@@ -881,16 +890,25 @@ function checkSemesterCredits(
...
+  let maxSeasonCredits;
+
+  //taking classes normally
+  if (status == "CLASSES") {
+    //minSeasonCredits is the minimum number of credits to not be under enrolled for the given season.
+    minSeasonCredits = seasonCreditTracker[season].seasonMin;
+
+    //maxSeasonCredits is the maximum number of credits to not be over enrolled for the given season.
+    maxSeasonCredits = seasonCreditTracker[season].seasonMax;
+  } else {
```
imanmoreira, line 906: Switch this to check if the status is coop, since right now this is including vacation semesters, and have it return no warnings if it's a vacation block

api/db/schema.rb
```diff
@@ -36,9 +36,11 @@
     t.string "username"
     t.integer "academic_year"
     t.integer "graduation_year"
+    t.string "major"
+    t.string "coop_cycle"
     t.index ["email"], name: "index_users_on_email", unique: true
     t.index ["reset_password_token"], name: "index_users_on_reset_password_token", unique: true
-    t.index ["username"], name: "index_users_on_username"
+    t.index ["username"], name: "index_users_on_username", unique: true
   end
```
imanmoreira, line 44: There shouldn't be any changes in this file.

PR #143, opened 2020-06-22
Transfer courses screen

liustanley:
Adds screen to mark completed courses as transfer credits. Transfer courses and completed schedule courses are separated such that transfer courses don't get added to the schedule, but are marked as completed in redux and the Sidebar.


imanmoreira (dismissed):
It looks good! I think there needs to be a little bit of cleanup that might be needed on the Transfer Credit screen code, and the AddClasses button to be added but otherwise it's all set.

frontend/src/Onboarding/TransferCoursesScreen.tsx
```diff
@@ -0,0 +1,363 @@
...
+import { AddClassModal } from "../components/AddClassModal";
+import { AddBlock } from "../components/ClassBlocks/AddBlock";
+
+const MainTitleText = styled.div`
+  font-size: 18px;
+  margin-left: 4px;
+  margin-top: 10px;
+  margin-bottom: 12px;
+  font-weight: 500;
+`;
```
andrewjleung, line 42: I noticed these styled divs for the most part closely resemble those also on the `CompletedCoursesScreen`, do we want to move these into a common file for reuse? Maybe `GenericOnboarding.tsx`?

frontend/src/Onboarding/GenericOnboarding.tsx
```diff
@@ -57,7 +58,7 @@ const theme = createMuiTheme({
   },
 });
 
-const steps = ["", ""];
+const steps = ["", "", ""];
```
andrewjleung, line 61: What is the purpose of steps? Why does it need to be initialized to an array of empty strings?

NilsBacke:
[Transfer Credit Pages](https://trello.com/c/tQPcYJcD/108-transfer-credit-pages)

frontend/src/Onboarding/TransferCoursesScreen.tsx
```diff
@@ -0,0 +1,363 @@
...
+import styled from "styled-components";
+import { fetchCourse } from "../api";
+import { NextButton } from "../components/common/NextButton";
+import { Link, withRouter, RouteComponentProps } from "react-router-dom";
+import { GenericOnboardingTemplate } from "./GenericOnboarding";
+import CheckBoxIcon from "@material-ui/icons/CheckBox";
+import CheckBoxOutlineBlankIcon from "@material-ui/icons/CheckBoxOutlineBlank";
+import {
+  Link as ButtonLink,
+  Collapse,
```
andrewjleung, line 28: Looks like both the `ButtonLink` alias and `Collapse` are both unused here

frontend/src/Onboarding/TransferCoursesScreen.tsx
```diff
@@ -0,0 +1,363 @@
...
+            minHeight: 300,
+            maxHeight: 300,
+            overflow: "-moz-scrollbars-vertical",
+            overflowY: "scroll",
+          }}
+          component={ScrollWrapper}
+        >
+          <Grid container justify="space-evenly">
+            <Grid key={0} item>
+              <Paper elevation={0} style={{ minWidth: 350, maxWidth: 400 }}>
```
NilsBacke, line 316: To add on to Andrew's comment, it seems that a lot of this is duplicated in the CompletedCoursesScreen as well. I think using a shared component would suit us well.

frontend/src/Onboarding/TransferCoursesScreen.tsx
```diff
@@ -0,0 +1,363 @@
...
+    this.setState({ modalVisible: true });
+  }
+
+  // Adds the "other courses" to the state in the form of a IRequiredCourse[][] so that they can be
+  // processed by renderAllCourses
+  addOtherCourses(courses: ScheduleCourse[]) {
+    let reqCourseMap = courses.map((course: ScheduleCourse) => [
+      {
+        type: "COURSE",
+        classId: +course.classId,
```
andrewjleung, line 153: Is this "+" doing anything here in particular?

frontend/src/Onboarding/TransferCoursesScreen.tsx
```diff
@@ -0,0 +1,363 @@
...
+        course.subject,
+        String(course.classId)
+      );
+
+      if (scheduleCourse) {
+        selectedCourses.push(scheduleCourse);
+      }
+    }
+
+    this.props.setCompletedCourses(completed);
```
andrewjleung, line 195: Just curious, why are completed courses being set here? It appears that we are taking the set completed requirements, filtering out those that are also taken as transfer credit, and re-setting the completed courses to that having already completed the completed courses step.

Is this to get rid of any overlap between what was selected as completed on the previous page and what was selected as transfer on this page?

frontend/src/Onboarding/TransferCoursesScreen.tsx
```diff
@@ -0,0 +1,363 @@
...
+        selectedRequirements: reqs,
+      });
+    }
+  }
+
+  /**
+   * Renders one course/courseset (if it contains labs/recitiations, and seperated)
+   * @param courses - Course pairings to be rendered
+   */
+  renderCourse(courses: IRequiredCourse[]) {
```
andrewjleung, line 224: This could also be shared with `CompletedCoursesScreen`.

andrewjleung (changes requested):
Looks good! We should look into sharing some components across the onboarding screens and I had a couple questions about little behaviors, but overall good work 👍

frontend/src/Onboarding/GenericOnboarding.tsx
imanmoreira: I just use it so I can use .map to create the Steps. It doesn't really need to be an empty array but it's just ease of use

frontend/src/Onboarding/TransferCoursesScreen.tsx
imanmoreira: Ooh fact, but we'll probably make that a separate tech debt ticket since this story is already kinda thicc

frontend/src/Onboarding/TransferCoursesScreen.tsx
imanmoreira: It converts the string into a number because in IRequiredCourse it's a number

frontend/src/Onboarding/TransferCoursesScreen.tsx
imanmoreira: Yeah it's to distinguish completed and trasfer credits, since completed courses populate into the schedule and transfer credits go into the transfer credit modal (not yet but soon)

PR #146, opened 2020-07-16
disable coop selection on semesters in first year

NilsBacke:


NilsBacke:
[Restrict Users from being on Coop before their Second year](https://trello.com/c/PlujsZyq/129-restrict-users-from-being-on-coop-before-their-second-year)

PR #148, opened 2020-07-16
reset coop cycle field on change of major in onboarding

NilsBacke:


NilsBacke:
[Reset the Coop-Cycle field in Onboarding and on the Major/Coop Modal when the user selects a new major so they can't accidentally select coop-cycles that don't exist on that major](https://trello.com/c/52Q6Iajz/125-reset-the-coop-cycle-field-in-onboarding-and-on-the-major-coop-modal-when-the-user-selects-a-new-major-so-they-cant-accidentally)

NilsBacke:
Changes made!

PR #149, opened 2020-07-17
go to sign up screen after initial onboarding if no major

NilsBacke:


NilsBacke:
[Ensure that onboarding goes to the signup step if the major is not filled in](https://trello.com/c/d3vkuNbe/124-ensure-that-onboarding-goes-to-the-signup-step-if-the-major-is-not-filled-in)

PR #151, opened 2020-07-17
redesign add class modal

NilsBacke:


NilsBacke:
[Redesign the add class modal](https://trello.com/c/Em0AFq2p/67-redesign-the-add-class-modal)

frontend/src/components/AddClassModal.tsx
```diff
@@ -220,20 +199,49 @@ export class AddClassModal extends React.Component<
...
+  renderClassIDTextField() {
+    return (
+      <TextField
+        id="outlined-basic"
+        label="Class ID"
+        variant="outlined"
+        value={this.state.formClassId}
+        onChange={this.handleClassIdChange.bind(this)}
+        placeholder="3500"
+        error={this.state.formClassId.length === 0}
```
imanmoreira, line 238: Right now it starts off in the error state. I'd say it doesn't really need to error at all, because we add error text when the request isn't filled out properly (same with the subject text field)

frontend/src/components/AddClassModal.tsx
```diff
@@ -220,20 +199,49 @@ export class AddClassModal extends React.Component<
...
+        style={{ marginTop: 12, marginBottom: 12, minWidth: 326 }}
+        color="secondary"
```
imanmoreira, line 240: I'd say don't use secondary since we don't use that in our color scheme anymore. You can leave it as default or use this color -> #EB5757 (which is the same as the NextButton) (same with the subject text field)

frontend/src/components/AddClassModal.tsx
```diff
@@ -260,40 +268,21 @@ export class AddClassModal extends React.Component<
...
+                ? "Add Class"
+                : "Add Classes"
+            }
             onClick={this.handleSubmit.bind(this)}
             disabled={this.state.queuedCourses.length === 0}
-          >
-            {this.state.queuedCourses.length === 1
-              ? "Add Class"
-              : "Add Classes"}
-          </SubmitButton>
```
imanmoreira, line 296: Not a huge deal but it feels like there's a lot of space above/beneath the Search button but almost no space under the Add Class button, so if it's not too difficult to balance out the space with those buttons that would be great

frontend/src/components/AddClassModal.tsx
```diff
@@ -260,40 +268,21 @@ export class AddClassModal extends React.Component<
...
-            </FormRow>
-          </form>
-          <SearchButton onClick={this.handleSearch.bind(this)}>
-            Search
-          </SearchButton>
+          <FormColumn>
+            {this.renderSubjectTextField()}
+            {this.renderClassIDTextField()}
+          </FormColumn>
+          <NextButton text="Search" onClick={this.handleSearch.bind(this)} />
```
imanmoreira, line 275: Nice refactor

PR #152, opened 2020-07-17
years are collapsible

NilsBacke:


NilsBacke:
[collapse all past years](https://trello.com/c/aBSkf9eh/19-collapse-all-past-years)

infrastructure/develop/run-servers.sh
```diff
@@ -6,7 +6,7 @@ if [[ ! " prod-build dev-build " =~ " $1 " ]]; then
 fi
 
 if [[ "dev-build" = "$1" ]]; then
-    docker-compose -f ../../api/docker-compose.yml up & cd ../../frontend && yarn start
+    docker-compose -f ../../api/docker-compose.yml up & cd ../../frontend && npm run start
```
imanmoreira, line 9: Why did you switch this to npm?

frontend/src/components/Year/Year.tsx
```diff
@@ -4,37 +4,80 @@ import { YearTop } from ".";
...
 
-export class Year extends React.Component<IYearProps> {
+export class Year extends React.Component<Props, State> {
+  constructor(props: Props) {
+    super(props);
+
+    const year = props.schedule.years[props.index];
+
+    this.state = {
+      expanded: !isYearInPast(year),
```
imanmoreira, line 41: It might be more accurate to use the academic year that the user enters (since if they're on their 3rd year, it's safe to assume you can collapse year 1 and 2)

frontend/src/utils/schedule-helpers.ts
```diff
@@ -189,3 +189,7 @@ export function getPositionOfYearInSchedule(
   }
   return index + 1;
 }
+
+export function isYearInPast(year: number): boolean {
+  return new Date().getFullYear() > year + 1;
```
imanmoreira, line 194: If you prefer to keep using the actual date, you should factor in the month of the year, since once it's September, the previous academic year should be collapsed.

infrastructure/develop/run-servers.sh
NilsBacke: Whoops this wasn't supposed to be included! Good catch

PR #155, opened 2020-07-23
fix deploy

ArunJ98:
Broke the deploy again and fixed it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,9 +14,9 @@
   "scripts": {
     "start": "cd infrastructure/develop && ./compose-build.sh && ./run-servers.sh dev-build",
     "prod-start": "cd infrastructure/develop && ./compose-build.sh && ./run-servers.sh prod-build",
-    "build": "cd frontend && CI=false react-scripts build",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -14,9 +14,9 @@
   "scripts": {
     "start": "cd infrastructure/develop && ./compose-build.sh && ./run-servers.sh dev-build",
     "prod-start": "cd infrastructure/develop && ./compose-build.sh && ./run-servers.sh prod-build",
-    "build": "cd frontend && CI=false react-scripts build",
+    "build": "cd frontend && CI=false react-app-rewired build",
     "test": "jest",
-    "eject": "cd frontend && react-scripts eject",
+    "eject": "cd frontend && react-app-rewired eject",
```
yasz24, line 19: I should have paid attention to this in the previous PR, but why do we need to use `react-app-rewired` ?

ArunJ98: its a library to do relative imports outside of src in Create-React-App without ejecting CRA and losing all those perks

yasz24: gotcha, makes sense

PR #160, opened 2020-09-24
Include Major and Coop Cycle in the Schedule State

andrewjleung:
## Why

https://trello.com/c/f1ogkMKP

Since user's should be able to choose majors and co-op cycles per schedule, we need to include these values within the schedule state, updating these values and pulling from them for schedule-specific actions.

## What
- Adds `major` and `coopCycle` fields to `ScheduleState` and `ScheduleStateSlice`
- In `Home`, `EditPlanPopper`, and `Sidebar`, pull values for `major` and `coopCycle` from the schedule state instead of the user state
- Add a new action and reducer case for setting the schedule major
- In `EditPlanPopper`, dispatch changes to `major` and `coopCycle` at the schedule level
- in `Sidebar`, pull major from the schedule state
- Fix issues within `RequirementSection` with setting state on an unmounted component by checking if the component is mounted before setting state.


frontend/src/components/Sidebar/Sidebar.tsx
```diff
@@ -64,7 +68,9 @@ const SidebarComponent: React.FC<Props> = ({ schedule, major }) => {
 
 const mapStateToProps = (state: AppState) => ({
   schedule: getScheduleFromState(state),
-  major: getDeclaredMajorFromState(state),
+  major: getMajors(state).find(
+    (major: Major) => major.name === getScheduleDataFromState(state).major
+  ),
```
andrewjleung, line 73: The schedule state only retains major as a string, while the sidebar requires a `Major`. This just retrieves the major from the stored `Major`s based upon what the string stored in the schedule is.

frontend/src/home/EditPlanPopper.tsx
```diff
@@ -139,24 +139,20 @@ export class EditPlanPopperComponent extends React.Component<
...
-
-    const plan = this.props.plans[this.props.major!.name].find(
+    const plan = this.props.plans[this.props.major].find(
       (p: Schedule) => planToString(p) === value
     );
 
     if (plan) {
-      this.props.setCoopCycle(plan);
+      const chosenCoopCycle = value === "None" ? "" : value;
+      this.props.setCoopCycle(chosenCoopCycle, plan);
```
andrewjleung, line 155: Do we want users to be able to unset coopCycle? In this case would I need to send `undefined` instead of an empty string if None was selected?

frontend/src/components/Sidebar/Sidebar.tsx
```diff
@@ -27,20 +27,28 @@ const MajorTitle = styled.p`
...
+    <Container>
+      <MajorTitle>No major selected</MajorTitle>
+    </Container>
+  );
+};
+
+const MajorSidebarComponent: React.FC<MajorSidebarProps> = ({
+  schedule,
+  major,
+}) => {
```
andrewjleung, line 51: Split out both sidebar cases into separate components.

frontend/src/components/Sidebar/Sidebar.tsx
```diff
@@ -62,9 +70,22 @@ const SidebarComponent: React.FC<Props> = ({ schedule, major }) => {
   );
 };
 
+const SidebarComponent: React.FC<SidebarProps> = ({ schedule, major }) => {
+  const majorObj: Major | undefined = useSelector<AppState, Major | undefined>(
+    state =>
+      state.majorState.majors.find((majorObj: Major) => majorObj.name === major)
+  );
```
andrewjleung, line 77: Since `major` exists in the schedule state as a string and the sidebar requires a `Major` object, we need to find the corresponding `Major` object from Redux.

frontend/src/components/Sidebar/Sidebar.tsx
```diff
@@ -62,9 +70,22 @@ const SidebarComponent: React.FC<Props> = ({ schedule, major }) => {
...
+    <MajorSidebarComponent schedule={schedule} major={majorObj} />
+  ) : (
+    <NoMajorSidebarComponent />
+  );
+};
+
 const mapStateToProps = (state: AppState) => ({
   schedule: getScheduleFromState(state),
-  major: getDeclaredMajorFromState(state),
+  major: getScheduleDataFromState(state).major,
```
andrewjleung, line 88: Fetch major string from schedule state instead of user.

frontend/src/state/actions/scheduleActions.ts
```diff
@@ -56,9 +57,14 @@ export const setScheduleAction = createAction(
   (schedule: Schedule) => ({ schedule })
 )();
 
+export const setScheduleMajor = createAction(
+  "schedule.SET_SCHEDULE_MAJOR",
+  (major?: Major) => ({ major })
+)();
```
andrewjleung, line 63: Add a new action for setting major at the schedule level.

frontend/src/state/reducers/scheduleReducer.ts
```diff
@@ -189,6 +194,18 @@ export const scheduleReducer = (
         draft.present.warnings = [];
         draft.present.courseWarnings = [];
 
+        // set the coop cycle
+        draft.present.coopCycle = coopCycle;
+
+        return draft;
+      }
+      case getType(setScheduleMajor): {
```
andrewjleung, line 202: Set reducer case for setting major at the schedule level.

frontend/src/home/EditPlanPopper.tsx
imanmoreira: Yeah we normally keep coopCycle as an optional field (coopCycle?: string is the same as coopCycle: string | undefined)

frontend/src/home/Home.tsx
```diff
@@ -185,8 +185,8 @@ interface ToastHomeProps {
 
 interface ReduxStoreHomeProps {
   schedule: DNDSchedule;
-  major?: Major;
-  planStr?: string;
+  major: string;
+  planStr: string;
```
imanmoreira, line 189: These should probably still be optional (and in other places as well)
This is causing some issues for majors that don't have options hooked up for coop cycles at the moment

frontend/src/home/Home.tsx
```diff
@@ -466,14 +467,15 @@ class HomeComponent extends React.Component<Props, HomeState> {
         name: `Schedule ${this.state.planCount + 1}`,
         link_sharing_enabled: this.props.linkSharing,
         schedule: this.props.schedule,
-        major: this.props.major ? this.props.major.name : "",
-        planString: this.props.planStr ? this.props.planStr : "None",
+        major: this.props.major,
+        planString: this.props.planStr,
```
imanmoreira, line 471: Same as above with this stuff

imanmoreira (changes requested):
Hey! Some minor changes I noticed but otherwise this is looking good!

frontend/src/components/Sidebar/Sidebar.tsx
NilsBacke: Looks like we have a helper for this in `plan-helpers.ts` called `findMajorFromName`

frontend/src/components/Sidebar/Sidebar.tsx
NilsBacke: Could we make a `getMajorFromState` function for this? Seems like we would use it a lot down the road.

frontend/src/home/EditPlanPopper.tsx
```diff
@@ -295,18 +291,19 @@ export class EditPlanPopperComponent extends React.Component<
 
 const mapStateToProps = (state: AppState) => ({
   schedule: getScheduleFromState(state),
-  planStr: getPlanStrFromState(state),
-  major: getDeclaredMajorFromState(state),
+  planStr: getScheduleDataFromState(state).coopCycle,
```
NilsBacke, line 294: And a `getCoopCycleFromState` would be sick too

NilsBacke:
[Include Major and Coop Cycle in the Schedule State](https://trello.com/c/f1ogkMKP/132-include-major-and-coop-cycle-in-the-schedule-state)

frontend/src/components/Sidebar/Sidebar.tsx
andrewjleung: That would make sense, I'll add that for major and coop cycle.

NilsBacke (approved):
LGTM to me, just need to fix conflicts

imanmoreira (approved):

PR #161, opened 2020-09-24
Add login logout buttons

justineluo:


justineluo:
[Add login/signout buttons](https://trello.com/c/eacoIHyB/138-add-login-signout-buttons)

frontend/src/Onboarding/SignupScreen.tsx
```diff
@@ -179,6 +179,7 @@ class SignupScreenComponent extends React.Component<Props, SignupScreenState> {
           });
         } else {
           const scheduleData: ScheduleSlice = this.props.getCurrentScheduleData();
+          console.log(response);
           createPlanForUser(response.user.id, response.user.token, {
```
NilsBacke, line 183: Can remove this log statement

frontend/src/Onboarding/SignupScreen.tsx
```diff
@@ -179,6 +179,7 @@ class SignupScreenComponent extends React.Component<Props, SignupScreenState> {
           });
         } else {
           const scheduleData: ScheduleSlice = this.props.getCurrentScheduleData();
+          console.log(response);
```
NilsBacke, line 182: Can remove this log statement

frontend/src/home/Home.tsx
```diff
@@ -236,6 +246,7 @@ class HomeComponent extends React.Component<Props, HomeState> {
     if (this.props.token && this.props.userId) {
       findAllPlansForUser(this.props.userId, this.props.token).then(
         (plans: IPlanData[]) => {
+          console.log(plans);
```
NilsBacke, line 249: Can remove this log too

frontend/src/home/Home.tsx
```diff
@@ -492,6 +503,11 @@ class HomeComponent extends React.Component<Props, HomeState> {
     this.props.setDNDSchedule(dndschedule);
   }
 
+  logOut() {
+    this.props.logOut();
+    this.updatePlan();
```
NilsBacke, line 508: I would call `updatePlan` before logout here

frontend/src/home/Home.tsx
```diff
@@ -583,6 +614,10 @@ const mapDispatchToProps = (dispatch: Dispatch) => ({
     dispatch(addNewSchedule(name, newSchedule)),
   updateActiveSchedule: (updatedSchedule: ScheduleSlice) =>
     dispatch(updateActiveSchedule(updatedSchedule)),
+  logOut: () => {
+    dispatch(resetUserAction());
+    dispatch(resetScheduleAction());
```
NilsBacke, line 619: It's generally not great practice to call dispatch twice in one prop. What I would recommend is to separate this out into two props: `resetUserAction` and `resetScheduleAction`, and then call both of those in the `logout` function.

frontend/src/state/reducers/scheduleReducer.ts
```diff
@@ -268,6 +269,10 @@ export const scheduleReducer = (
         draft.present.warnings = namedSchedule.warnings;
         return draft;
       }
+      case getType(resetScheduleAction): {
+        draft = initialState;
+        return draft;
+      }
     }
```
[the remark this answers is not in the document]
imanmoreira: Another thing, you can actually reuse the resetUserAction here instead of making a separate one for Schedule (you can have the ScheduleAction track UserActions)

frontend/src/Onboarding/Onboarding.tsx
```diff
@@ -123,7 +131,16 @@ class OnboardingComponent extends React.Component<OnboardingProps> {
 
     return (
       <Container>
-        <Header></Header>
+        <Header>
+          <LoginLink
+            to={{ pathname: "/login", state: { userData: {} } }}
```
imanmoreira, line 136: What does the state value do here? (I'm more just curious)

imanmoreira (changes requested):

frontend/src/Onboarding/LoginScreen.tsx
```diff
@@ -269,7 +268,7 @@ class LoginScreenComponent extends React.Component<Props, LoginScreenState> {
 
         <Subtitle>
           New here? Sign up{" "}
-          <Link style={{ color: "#EB5757" }} to="/signup">
+          <Link style={{ color: "#EB5757" }} to="/onboarding">
```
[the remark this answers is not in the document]
NilsBacke: Iman brings up a good point, I think we'll have to pass in a parameter to the Login screen to determine which route we redirect to. The issue is if a user continues as a guest, then from the home screen presses "Login", then they click the sign in link, they were never presented with the initial 4 onboarding questions in `/onboarding`.

frontend/src/Onboarding/Onboarding.tsx
NilsBacke: I believe it was just copied and pasted from somewhere else, so I think we can remove `state` here

frontend/src/Onboarding/LoginScreen.tsx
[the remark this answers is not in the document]
justineluo: When the user attempts to sign up via the "login button" on the welcome page without going through onboarding, the serve responds with 422 error, since the payload is missing some information that is collected via onboarding page. When the user goes through "get started" they land on the signup page, so unless they toggle between sign up and login, they probably wont have to go through onboarding again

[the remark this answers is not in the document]
justineluo: why would we disable "continue as guest" from welcome page? do you mean via the login button?

frontend/src/Onboarding/CompletedCoursesScreen.tsx
```diff
@@ -343,7 +347,7 @@ class CompletedCoursesComponent extends Component<Props, State> {
           handleSubmit={courses => this.addOtherCourses(courses)}
         ></AddClassModal>
         <Link
-          to={"/signup"}
+          to={fromOnBoarding ? "/home" : "/signup"}
```
NilsBacke, line 350: I think this should be the other way around? Or am I totally off on that

NilsBacke:
One thing worth discussing: on the home page if you're in guest mode, I think it's better to have a "Sign Up" button as opposed to a "Login" button.

frontend/src/home/Home.tsx
```diff
@@ -492,6 +511,11 @@ class HomeComponent extends React.Component<Props, HomeState> {
     this.props.setDNDSchedule(dndschedule);
   }
 
+  logOut() {
+    this.updatePlan();
+    this.props.logOut();
```
NilsBacke, line 516: I'm getting an error on logout because we're trying to call `this.props.updateActiveSchedule` after the logOut has been run. What I would recommend is to convert `this.updatePlan()` to an async function and use `await` instead of `.then`. That way in `logOut()` we can do 
```suggestion
    this.updatePlan().then(() => {
      this.props.logOut();
    });
```

frontend/src/Onboarding/CompletedCoursesScreen.tsx
```diff
@@ -343,7 +347,9 @@ class CompletedCoursesComponent extends Component<Props, State> {
           handleSubmit={courses => this.addOtherCourses(courses)}
         ></AddClassModal>
         <Link
-          to={"/signup"}
+          to={{
+            pathname: fromOnBoardingGuest ? "/home" : "/signup",
+          }}
```
imanmoreira, line 352: I really like this

frontend/src/Onboarding/LoginScreen.tsx
```diff
@@ -268,11 +272,22 @@ class LoginScreenComponent extends React.Component<Props, LoginScreenState> {
...
+          >
             here
           </Link>{" "}
           or{" "}
-          <Link style={{ color: "#EB5757" }} to="/home">
+          <Link
+            style={{ color: "#EB5757" }}
+            to={{
+              pathname: fromOnBoarding ? "/onboarding" : "/home",
+              state: { fromOnBoardingGuest: fromOnBoarding },
```
imanmoreira, line 288: Damn look at you using the state 💯

frontend/src/Onboarding/OnboardingInfoScreen.tsx
```diff
@@ -354,7 +367,12 @@ class OnboardingScreenComponent extends React.Component<
 
           {textFieldStr.length !== 0 && !!year && !!gradYear ? (
             <Link
-              to={!!this.state.major ? "/completedCourses" : "/signup"}
+              to={{
+                pathname: !!this.state.major
+                  ? "/completedCourses"
+                  : this.onBoardLink(fromOnBoardingGuest),
```
imanmoreira, line 373: You can actually chain terenary operators like this
!!this.state.major  ? "/completedCourses" :  fromOnBoardingGuest ? "/home" : "/signup"

imanmoreira (changes requested):

imanmoreira (approved):

PR #169, opened 2020-10-08
Add catalog year

William-Cunningham:
adds catalog year field and changes onboarding/signup screen to account for this.


imanmoreira (changes requested):
The signup screen error we discussed

frontend/src/Onboarding/OnboardingInfoScreen.tsx
```diff
@@ -316,11 +339,48 @@ class OnboardingScreenComponent extends React.Component<
...
+    //need to make chanegs to options to filter out repeat catalog years
+    let majorSet = [
+      ...Array.from(
+        new Set(this.props.majors.map(maj => maj.yearVersion.toString()))
+      ),
+    ];
+    return (
+      <Tooltip
+        title="Catalog Year refers to the year your major credits are associated to. This is usually the year you declared your Major."
+        placement="top"
```
NilsBacke, line 355: nice tooltip :)

frontend/src/Onboarding/OnboardingInfoScreen.tsx
```diff
@@ -180,11 +200,14 @@ class OnboardingScreenComponent extends React.Component<
    * Renders the major drop down
    */
   renderMajorDropDown() {
+    let majorNames = this.props.majors.filter(
+      major => major.yearVersion === this.state.catalogYear
```
NilsBacke, line 204: Do all of our majors have a yearVersion? It might be safer to show majors with no year version and those with the same year version.

Also, is there a major for every year version? Is there the case where a major has a year version, let's say 2018, and there does not exist a 2019 year version for that same major. The user selects a catalog year of 2019. What happens then? 

I'm not too familiar with how the majors work and such but just want to make sure that we handle all of these cases (if they are possible)

NilsBacke (approved):

frontend/src/Onboarding/OnboardingInfoScreen.tsx
```diff
@@ -312,6 +337,42 @@ class OnboardingScreenComponent extends React.Component<
...
+    return (
+      <Tooltip
+        title="Catalog Year refers to the year your major credits are associated to. This is usually the year you declared your Major."
+        placement="top"
+      >
+        <Autocomplete
+          style={{ width: 326, marginBottom: marginSpace }}
+          disableListWrap
+          //map by year then filter for repeats (also do for major dropdown)
+          //do I need to provide menu item options like in graduation year ^^^
```
justineluo, line 359: I think you mean to remove this comment :)

frontend/src/home/EditPlanPopper.tsx
```diff
@@ -299,12 +341,20 @@ const mapStateToProps = (state: AppState) => ({
...
+  isLoggedIn: isUserLoggedIn(state),
 });
 
 const mapDispatchToProps = (dispatch: Dispatch) => ({
   setCoopCycle: (coopCycle: string, schedule?: Schedule) =>
     dispatch(setCoopCycle(coopCycle, schedule)),
   setSchedule: (schedule: Schedule) => dispatch(setScheduleAction(schedule)),
+  //change this to use schedule action instead of user
+  //edit popper clearing bug
+  //adding catalog year to backend schedule model
```
justineluo, line 355: You'll want to remove these comments too

frontend/src/home/EditPlanPopper.tsx
```diff
@@ -184,25 +199,49 @@ export class EditPlanPopperComponent extends React.Component<
...
   }
 
+  renderCatalogYearDropdown() {
+    let catalogYears = [
+      ...Array.from(
+        new Set(this.props.majors.map(maj => maj.yearVersion.toString()))
+      ),
+    ];
+    catalogYears.push("2020");
+    catalogYears.push("2021");
```
justineluo, line 224: I think we had these pushes for testing - we'll want to remove these

frontend/src/home/Home.tsx
```diff
@@ -281,7 +283,8 @@ class HomeComponent extends React.Component<Props, HomeState> {
         (plans: IPlanData[]) => {
           // Once multiple plans are supported, this can be changed to the last used plan
           let plan: IPlanData = plans[0];
-
+          console.log(plan);
+          console.log(plans);
```
justineluo, line 287: console log

frontend/src/profile/Profile.tsx
```diff
@@ -306,6 +352,8 @@ export const ProfileComponent: React.FC = (props: any) => {
   const [advisor, setAdvisor] = useState("");
   const [email, setEmail] = useState(props.email);
   const [coop, setCoop] = useState(props.coop);
+  //ask about hooks
```
imanmoreira, line 355: You can remove this

frontend/src/Onboarding/OnboardingInfoScreen.tsx
imanmoreira: Yup! All majors will have a year version, and majors can not exist on certain years (ex information science no longer exists and several data science combined majors have been added). I'll let Nathan or Jmoney respond about the code behavior but just for the context

imanmoreira (dismissed):

NilsBacke (approved):

frontend/src/Onboarding/OnboardingInfoScreen.tsx
justineluo: If there is no year version for that major, it will not show up in the drop down. To go off of Iman's example, if you pick 2020 as the catalog year, you won't see Information Science since it doesn't exist anymore and will be filtered out. I think it would make sense for this to be the expected behavior.

api/db/migrate/20201115193346_add_catalog_year_to_plans.rb
```diff
@@ -0,0 +1,5 @@
+class AddCatalogYearToPlans < ActiveRecord::Migration[6.0]
+  def change
+    add_column :plans, :catalog_year, :integer, default: 2018
```
NilsBacke, line 3: Wondering if this should default to the current year when a user makes a plan. In this case we don't need a default here

NilsBacke (approved):

PR #177, opened 2020-11-07
Transfer credit holder

justineluo:
Render transfer credit holder at the bottom on home screen. Based on discussion with team, no need to make classes from transfer holder draggable (may need further discussion + separate PR for that) 

Notes: 
building off of transfer-credit changes 
(pls test login/logout)


andrewjleung:
I recognize some of these changes from previous work that should already be in master, are these correct? Also should we only review the changes for the transfer credit holder?

justineluo:
oh right transfer-credit is behind master but transfer-credit-holder has some changes from master. I'll merge master into transfer-credit so we can see my changes better

justineluo:
[Create the transfer credit holder](https://trello.com/c/YLgfzT3E/135-create-the-transfer-credit-holder)

frontend/src/home/Home.tsx
```diff
@@ -452,6 +461,37 @@ class HomeComponent extends React.Component<Props, HomeState> {
     }
   }
 
+  renderTransfer() {
+    // If a user is currently logged in, wait until plans are fetched to render
+    if (this.props.token && this.props.userId) {
```
andrewjleung, line 466: Looks like this component takes in `isLoggedIn` from state, so this can be replaced with `this.props.isLoggedIn` just to make things a little more readable 😃

frontend/src/state/reducers/scheduleReducer.ts
```diff
@@ -134,6 +142,17 @@ export const scheduleReducer = (
         draft.present = JSON.parse(JSON.stringify(draft.past));
         return draft;
       }
+      case getType(removeTransferClassAction): {
+        const { course } = action.payload;
+        // save prev state with a deep copy
+        draft.past = JSON.parse(JSON.stringify(draft.present));
```
andrewjleung, line 148: We may want to look into a different method of deep copying here, I'm not sure if it is safe to use JSON stringify/parse on a `ScheduleStateSlice` without potentially losing data since it contains `closedYears` which is a `Set`.

https://stackoverflow.com/questions/122102/what-is-the-most-efficient-way-to-deep-clone-an-object-in-javascript

frontend/src/components/TransferCreditHolder.tsx
```diff
@@ -0,0 +1,161 @@
...
+
+const Wrapper = styled.div`
+  display: flex;
+  flex-direction: row;
+  flex-wrap: wrap;
+  width: 100%;
+  height: 100%;
+`;
+
+class TransferCreditsComponent extends React.Component<
```
andrewjleung, line 71: Nice!!!!

andrewjleung (dismissed):
Just one change for readability and also some deep copying business, otherwise this looks great! Nice job @justineluo !!!

frontend/src/components/TransferCreditHolder.tsx
```diff
@@ -0,0 +1,161 @@
...
+        </HolderBody>
+      </div>
+    );
+  }
+}
+
+const mapStateToProps = (
+  state: AppState,
+  ownProps: TransferCreditsProps
+) => ({});
```
NilsBacke, line 144: We can remove this here as it's unused (same with the relevant Props definition)

frontend/src/state/reducers/scheduleReducer.ts
justineluo: Good catch!! I'm going to keep stringify for now since I'm having issue with cloneDeep. Iman's going to make a ticket for it in the backlog :)

frontend/src/home/Home.tsx
```diff
@@ -453,6 +462,37 @@ class HomeComponent extends React.Component<Props, HomeState> {
     }
   }
 
+  renderTransfer() {
+    // If a user is currently logged in, wait until plans are fetched to render
+    if (this.props.isLoggedIn) {
```
imanmoreira, line 467: I think we talked about this refactor, but this case can be simplified to this.props.isLoggedIn && ! this.state.fetchedPlan 
-> SpinnerWrapper, else render transferCredits

frontend/src/components/TransferCreditHolder.tsx
```diff
@@ -0,0 +1,147 @@
...
+          <AddBlock onClick={this.showModal.bind(this)} />
+          <AddClass
+            visible={modalVisible}
+            handleClose={this.hideModal.bind(this)}
+            handleSubmit={(courses: ScheduleCourse[]) => {
+              // Add the given courses through redux
+              this.props.handleAddClasses(courses);
+            }}
+          ></AddClass>
+        </HolderBody>
```
imanmoreira, line 131: I think we talked about this as well, but add in that undo snackbar for when the user deleted a class 
(See Semester Block)

imanmoreira (changes requested):
I think everything else looks good. I thought about ticket breakdown and decided the backend additions we talked about could be a separate ticket, since we don't want to hold this off from being merged in when it's in a good state right now.

imanmoreira (approved):

PR #189, opened 2020-11-12
add annotate gem and annotate models

NilsBacke:
This is so it's a lot easier to know which fields are on each model


NilsBacke:
[Add annotate gem and annotate models](https://trello.com/c/asR4Hjvk/174-add-annotate-gem-and-annotate-models)

imanmoreira (approved):

PR #192, opened 2020-11-12
Add Template plans

NilsBacke:


NilsBacke:
[Create template plans model and add fields onto existing plans model](https://trello.com/c/v4EVkWzP/176-create-template-plans-model-and-add-fields-onto-existing-plans-model)

imanmoreira (approved):

imanmoreira (commented):
I think you have it set to merge into annotate so you might want to switch it to master

api/db/schema.rb
```diff
@@ -27,9 +32,22 @@
...
+    t.boolean "is_currently_being_edited", default: false, null: false
     t.index ["user_id"], name: "index_plans_on_user_id"
   end
 
+  create_table "template_plans", force: :cascade do |t|
+    t.string "name", null: false
+    t.integer "catalog_year", null: false
+    t.json "schedule"
+    t.string "major", null: false
+    t.string "plan_string", null: false
```
imanmoreira, line 44: Let's switch plan_string to coop_cycle, since now we have "plans" which are schedules that a user has

imanmoreira (changes requested):

imanmoreira (approved):

PR #195, opened 2020-11-15
Fix Deep Copying Error caused by Closed years

imanmoreira:
Switch redux props for closed years to be an array instead of a set to avoid deep copying errors


imanmoreira:
[Modify Deep Copying in Redux](https://trello.com/c/DKH6YokI/178-modify-deep-copying-in-redux)

ArunJ98 (approved):

PR #203, opened 2020-11-22
Set up Redux Store for Advisors

andrewjleung:
https://trello.com/c/sxNN1KLs

## Why

We need to draft out the design of the Redux store for the advisor work flow as we start getting into implementing it.

## What

- Create an `AdvisorState` interface containing basic session data along a list of students for use in both the students tab and when searching students to assign a template to
- Add the `advisorReducer` along with advisor actions

## Notes

We are currently settling with not persisting data like notifications or templates within the Redux state and requesting them as needed since they are only used in a single location each. Students on the other hand will be cached in Redux after the first request so it can be reused across both the students tab and when assigning templates to students.

Please let us know if there are any concerns with this approach!

### DIFF
--- a/frontend/src/state/actions/advisorActions.ts
+++ b/frontend/src/state/actions/advisorActions.ts
@@ -1,0 +1,32 @@
+import { createAction } from "typesafe-actions";
+import { Student } from "../reducers/advisorReducer";
+
+export const setEmail = createAction("advisor/SET_EMAIL", (email: string) => ({
+  email,
+}))();
+
+export const setName = createAction("advisor/SET_NAME", (name: string) => ({
+  name,
+}))();
+
+export const setToken = createAction("advisor/SET_TOKEN", (token: string) => ({
+  token,
+}))();
+
+export const setUserId = createAction(
+  "advisor/SET_USER_ID",
+  (userId: number) => ({
+    userId,
+  })
+)();
+
+export const setImage = createAction("advisor/SET_IMAGE", (image: string) => ({
+  image,
+}))();
+
+export const setStudents = createAction(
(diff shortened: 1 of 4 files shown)


frontend/src/state/reducers/advisorReducer.ts
```diff
@@ -0,0 +1,73 @@
...
+  setName,
+  setStudents,
+  setToken,
+  setUserId,
+} from "../actions/advisorActions";
+
+export interface Student {
+  readonly email: string;
+  readonly name: string;
+}
```
andrewjleung, line 16: This simply contains information needed to display a list of users to the advisor, where the student's email uniquely defines them. 

The idea is that when an advisor selects a student, this email is used to fire off a new request to retrieve the rest of that student's information.

frontend/src/state/actions/advisorActions.ts
```diff
@@ -0,0 +1,32 @@
+import { createAction } from "typesafe-actions";
+import { Student } from "../reducers/advisorReducer";
+
+export const setEmail = createAction("advisor/SET_EMAIL", (email: string) => ({
```
NilsBacke, line 4: We might run into some issues if we use this action in a file that also uses the equivalent "user" action. Maybe `setAdvisorEmail` is better, but not sure if we'll run into this issue.

NilsBacke:
This is great. We might want to rename the existing user actions and reducer to student actions and reducer (in a future PR)

NilsBacke (approved):

NilsBacke:
[Start Setting up Redux Store for Advisors](https://trello.com/c/sxNN1KLs/169-start-setting-up-redux-store-for-advisors)

PR #204, opened 2020-11-22
advisor get student plan without token 

justineluo:


justineluo:
[Change all plans endpoint so that no token is required if you are an advisor](https://trello.com/c/KtdC0Glm/196-change-all-plans-endpoint-so-that-no-token-is-required-if-you-are-an-advisor)

imanmoreira (approved):

PR #210, opened 2020-11-29
Advisor Manage Students - Connect with backend

ArunJ98:


frontend/src/advising/ManageStudents.tsx
```diff
@@ -29,63 +35,151 @@ const StudentListContainer = styled.div`
...
+              username={student.username}
+              // nuid={student.nuid}
+              email={student.email}
+              key={student.username + Math.random()}
+            />
+          ))
+        )}
+        {!isLoading ? (
+          isLastPage ? (
+            <NoMoreStudents>No more students</NoMoreStudents>
```
NilsBacke, line 160: Maybe instead of displaying this, we could just not show the "Load More" button

ArunJ98: I wanted to let the user know that we got no results back. If the user hits Load more and then the button just disappears but nothing came back, it's more ambiguous what happened. Either way - seems like we might want to do infinite scroll here so we can experiment around and see what's best.

NilsBacke:
Looks like you can remove adding the index on username as I never removed it in the auth PR. Other than that looks good to go 👍

NilsBacke (approved):

PR #212, opened 2020-12-01
remove redux persist and get auth token from cookie

NilsBacke:
It's a security risk to save our JWT user token in local storage, it's better to save it in a cookie. Also, we can just remove redux-persist altogether, as we should be relying on the backend and not have our redux store get out of sync with the backend.

Also the logout is fully functional (from what I can tell), and we properly redirect users if they are/aren't logged in to the appropriate page.

Also we now have to always load user data on app load if it does not already exist in our redux store.


NilsBacke:
[Remove redux-persist from persisting our redux store and use token from cookie](https://trello.com/c/TzMKvugT/205-remove-redux-persist-from-persisting-our-redux-store-and-use-token-from-cookie)

ArunJ98 (approved):
nice!

PR #220, opened 2020-12-03
Refactor redux store to support multiple plans

NilsBacke:
Overview of changes:
Remove scheduleReducer and scheduleActions
Replace schedulesReducer and schedulesActions with userPlansReducer and userPlansActions
Rename all state getters to be more explicit and end in "FromState"
Rename all actions to be more explicit and end in "Action"

Add completed courses, transfer courses, and exam credits to the user store (as they are specific to a user, not a plan)


common/types.ts
```diff
@@ -260,7 +260,6 @@ export interface Schedule {
   yearMap: {
     [key: number]: ScheduleYear;
   };
-  id: string;
```
NilsBacke, line 263: We don't need the id field on a schedule anymore

frontend/src/Onboarding/LoginScreen.ts
```diff
@@ -0,0 +1,350 @@
+export {};
```
NilsBacke, line 1: Going to comment this whole file out for now, as we won't have to worry about it (until we expand to other colleges)

frontend/src/Onboarding/Onboarding.tsx
```diff
@@ -255,14 +251,3 @@ class OnboardingComponent extends React.Component<Props> {
     );
   }
 }
-
-const mapStateToProps = (state: AppState) => ({
-  fullName: getFullNameFromState(state),
-});
```
NilsBacke, line 261: Unused variables

frontend/src/Onboarding/OnboardingInfoScreen.tsx
```diff
@@ -337,21 +330,21 @@ const mapDispatchToProps = (dispatch: Dispatch) => ({
     dispatch(setAcademicYearAction(academicYear)),
   setGraduationYear: (academicYear: number) =>
     dispatch(setGraduationYearAction(academicYear)),
-  setMajor: (major?: Major) => dispatch(setDeclaredMajorAction(major)),
-  setCoopCycle: (coopCycle: string, plan: Schedule) =>
-    dispatch(setCoopCycle(coopCycle, plan)),
+  setMajor: (major?: Major) => dispatch(setUserMajorAction(major?.name || '')),
```
NilsBacke, line 333: We only store the major name in redux (to stay consistent with the backend), and we can always use `findMajorFromName` if we need the object

frontend/src/Onboarding/RedirectScreen.tsx
```diff
@@ -90,17 +80,7 @@ export const RedirectScreen: React.FC<Props> = ({ redirectUrl }) => {
...
-          dispatch(setAcademicYearAction(response.user.academicYear));
-          const maj = findMajorFromName(response.user.major, majors);
-          if (maj) {
-            dispatch(setDeclaredMajorAction(maj));
-          }
-          dispatch(setUserIdAction(response.user.id));
-          dispatch(setEmailAction(response.user.email));
-          dispatch(setUserCoopCycleAction(response.user.coopCycle));
-          dispatch(setIsAdvisorAction(response.user.isAdvisor));
+          dispatch(setUserAction(response.user));
```
NilsBacke, line 83: We can just set the whole user object instead of setting each field individually

NilsBacke: Keep in mind this requires that the fields in the user object in redux must match the user object returned from the backend

frontend/src/components/TransferCreditHolder.tsx
```diff
@@ -190,7 +185,6 @@ const mapDispatchToProps = (dispatch: Dispatch) => ({
     dispatch(addTransferClassAction(courses)),
   onDeleteClass: (course: ScheduleCourse) =>
     dispatch(removeTransferClassAction(course)),
-  onUndoDeleteClass: () => dispatch(undoRemoveClassAction()),
```
NilsBacke, line 193: I created another trello card to re-implement undoing deleting a transfer class

frontend/src/home/Home.tsx
```diff
@@ -275,32 +243,17 @@ class HomeComponent extends React.Component<Props, HomeState> {
...
-          });
-        }
-      );
-    } else if (this.props.activeSchedule) {
-      this.setState({ fetchedPlan: true });
-    }
+    const token = getAuthToken();
+    findAllPlansForUser(this.props.userId!, token).then(
+      (plans: IPlanData[]) => {
+        this.props.setUserPlans(plans, this.props.academicYear);
```
NilsBacke, line 249: This also sets the active plan

frontend/src/home/Home.tsx
```diff
@@ -313,23 +266,25 @@ class HomeComponent extends React.Component<Props, HomeState> {
...
-      );
-      if (!this.props.closedYears.has(yearIdx)) {
-        numVisibleWarnings++;
-        if (numVisibleWarnings <= 5) {
-          // add new toasts
-          this.props.addToast(w.message, {
-            appearance: "warning",
-            autoDismiss: true,
-          });
+    if (this.props.activePlan) {
```
NilsBacke, line 269: Only do this is we have an active plan. There is a time while the user plans are loading in the home screen that there is not an active plan, so we only want to render things that do not rely on an active plan while the user plans are loading

frontend/src/home/Home.tsx
```diff
@@ -338,20 +293,20 @@ class HomeComponent extends React.Component<Props, HomeState> {
     // if drag is coming from the sidebar
     if (isNaN(Number(source.droppableId))) {
       addCourseFromSidebar(
-        this.props.schedule,
+        this.props.activePlan!.schedule,
```
NilsBacke, line 296: In order to be able to add a course from sidebar, there must be an active plan, so it's okay to assert here.

frontend/src/home/Home.tsx
```diff
@@ -476,82 +437,25 @@ class HomeComponent extends React.Component<Props, HomeState> {
...
       }
     });
   }
 
-  /**
-   * If a user is currently logged in, saves the current plan under this user.
-   * Only supports updating a user's singular plan, can be modified later to
-   * update a specific plan.
-   */
-  savePlan() {
```
NilsBacke, line 524: Remove this unused function

frontend/src/home/Home.tsx
```diff
@@ -564,6 +468,20 @@ class HomeComponent extends React.Component<Props, HomeState> {
   };
 
   render() {
+    if (!this.state.fetchedPlan) {
```
NilsBacke, line 471: To air on the safe side, just render a loading indicator if currently fetching user plans

frontend/src/profile/Profile.tsx
```diff
@@ -140,23 +136,23 @@ interface ChangePasswordProps {
...
-          value={props.name}
-          onChange={e => props.setName(e.target.value)}
-          placeholder="John Smith"
-        />
-      )}
-      {!props.isEdit && <ItemEntry> {props.name} </ItemEntry>}
-    </ProfileEntryContainer>
-  );
-};
+// const ProfileName = (props: ProfileNameProps) => {
```
NilsBacke, line 139: Going to re-implement this in a separate PR, as right now we don't even show the Profile page. Right now we want to show their full name, but not allow them to edit it.

NilsBacke:
[Get ScheduleAction/Reducer renamed so code is readable and clean up reducers (a lot)](https://trello.com/c/LuJKx21c/151-get-scheduleaction-reducer-renamed-so-code-is-readable-and-clean-up-reducers-a-lot)

frontend/src/Onboarding/RedirectScreen.tsx
ArunJ98: don't think this will work for transfer/completed courses. What happens if we pass the response and it doesn't match, will it just not set those fields or will it error?

frontend/src/profile/Profile.tsx
ArunJ98: yea we might want to rethink the UI/purpose of this page, probably not super high priority right now

frontend/src/home/Home.tsx
ArunJ98: how do we determine which one is the active plan?

frontend/src/home/Home.tsx
ArunJ98: nice

frontend/src/state/index.ts
```diff
@@ -176,66 +137,87 @@ export const getTakenCredits = (state: AppState): number =>
...
-export const getIsAdvisorFromState = (state: AppState): boolean | undefined => {
-  return state.user.isAdvisor;
+export const getIsAdvisorFromState = (state: AppState) => {
+  return getUserFromState(state).isAdvisor;
 };
+
+/*
+ * SAFE GET FUNCTIONS
+ * Should only need to use these where there is not guaranteed to have an active plan (like the home screen)
+ * TODO: look into if a separate component (prior to getting to the HomeComponent) to load plan data would be worth it
```
ArunJ98, line 209: What if we just loaded all plan info on the loading screen on the redirect page. Once we get the user, we can get all necessary plan info/etc... We can do this in a separate PR but I think this would simplify the flow.

NilsBacke: I did consider this when coming up with solutions, it seemed like a lot of changes at the time. I do think that your idea is the best though, basically for the reason that you shouldn't be able to go to the HomeScreen without an active plan (which logically makes sense).

NilsBacke: I added a card for this in Trello

NilsBacke:
We determine which plan is active by just choosing the first plan for now (we do this in the reducer for setUserPlansAction). We have a card in trello to add a `lastViewed` timestamp field on a plan, and the most recently viewed plan will be made the active plan.

NilsBacke:
"don't think this will work for transfer/completed courses. What happens if we pass the response and it doesn't match, will it just not set those fields or will it error?"

It won't error, it just won't save correctly. 

For completed/transfer courses could we just make that part of the PR you're working on? Right now this is the same functionality as in master right now, where we don't handle completed/transfer courses from the backend.

ArunJ98 (approved):

NilsBacke:
[Fully support multiple plans in the frontend/backend](https://trello.com/c/xTtClswf/213-fully-support-multiple-plans-in-the-frontend-backend)

PR #221, opened 2020-12-04
set active plan to the last viewed plan

NilsBacke:


NilsBacke:
[Make it so when a user logs in, the active plan is the plan they last looked at](https://trello.com/c/u5dKZxAq/211-make-it-so-when-a-user-logs-in-the-active-plan-is-the-plan-they-last-looked-at)

api/app/models/user.rb
```diff
@@ -2,26 +2,28 @@
...
+#  image_url              :string
+#  is_advisor             :boolean          default(FALSE), not null
+#  major                  :string
+#  remember_created_at    :datetime
+#  reset_password_sent_at :datetime
+#  reset_password_token   :string
+#  username               :string
+#  created_at             :datetime         not null
+#  updated_at             :datetime         not null
+#  nu_id                  :string
```
NilsBacke, line 20: These schema annotations reflect what our migrations build, so these should be in master. If these comments change when you don't expect them to, try doing a rails db:drop db:create db:migrate to your database.

NilsBacke:
[Remove jwt token from current user endpoint](https://trello.com/c/Xwt8AUvj/228-remove-jwt-token-from-current-user-endpoint)

ArunJ98 (approved):

PR #226, opened 2020-12-04
use shallow equal

NilsBacke:
ShallowEqual is necessary for proper rerenders when we use `useSelector` in this way (returning one object with everything), as opposed to one `useSelector` call per field we want from redux.


NilsBacke:
[Use shallowEqual for all useSelector instances](https://trello.com/c/9XazuyBv/238-use-shallowequal-for-all-useselector-instances)

imanmoreira (approved):

PR #227, opened 2020-12-04
dont use empty string as null value

NilsBacke:


NilsBacke:
[Make "missing" string values be undefined, not empty string](https://trello.com/c/kUYMzW3B/231-make-missing-string-values-be-undefined-not-empty-string)

frontend/src/Onboarding/RedirectScreen.tsx
```diff
@@ -55,14 +55,11 @@ interface Props {
 
 export const RedirectScreen: React.FC<Props> = ({ redirectUrl }) => {
   const dispatch = useDispatch();
-  const { academicYear, graduationYear, isAdvisor } = useSelector(
-    (state: AppState) => ({
-      academicYear: getAcademicYearFromState(state),
-      graduationYear: getGraduationYearFromState(state),
-      isAdvisor: getIsAdvisorFromState(state),
-    })
-  );
```
NilsBacke, line 64: Getting these values here would error, as there is no user in our redux store at this point in the flow

frontend/src/state/reducers/userReducer.ts
```diff
@@ -18,34 +18,14 @@ import {
...
-  sumCreditsFromList,
-  numToTerm,
-  getNextTerm,
-  produceWarnings,
-} from "../../utils";
 
 export interface UserState {
-  user: IUserData;
+  user?: IUserData;
   completedRequirements: IRequiredCourse[]; // only used in onboarding flow
```
NilsBacke, line 24: This pattern better reflects how we're using this reducer in conjunction with our api

imanmoreira (approved):

ArunJ98:
Merging this as it affects the PR I'm working on

PR #229, opened 2020-12-05
batch consecutive redux actions

NilsBacke:


NilsBacke:
[Batch redux actions that are called at the same time](https://trello.com/c/F3te4j21/236-batch-redux-actions-that-are-called-at-the-same-time)

imanmoreira (approved):

PR #230, opened 2020-12-05
Load everything needed for home before we get there.

ArunJ98:


ArunJ98:
[Move loading of active plan to RedirectScreen](https://trello.com/c/6EjJZWz3/227-move-loading-of-active-plan-to-redirectscreen)

frontend/src/home/Home.tsx
```diff
@@ -457,30 +388,16 @@ class HomeComponent extends React.Component<Props, HomeState> {
...
   logOut = async () => {
     await this.updatePlan(false);
-    this.props.logOut();
     removeAuthTokenFromCookies();
 
     alert(
       "Your plan has been updated and you have been logged out. You will be redirected to the welcome screen."
     );
     this.props.history.push("/");
+    this.props.logOut();
```
ArunJ98, line 397: We need to do this after we dismount Home.

frontend/src/Onboarding/RedirectScreen.tsx
```diff
@@ -99,31 +70,25 @@ export const RedirectScreen: React.FC<Props> = ({ redirectUrl }) => {
               setIsLoading(false); // this update must come last, to make sure other state variables are correctly set before we redirect
             });
           })
-          .catch(e => console.log(e));
+          .catch(e => {
+            // TODO: Log error to some service like rollbar
+            console.log(e);
+            setIsError(true);
```
NilsBacke, line 76: We should also `setIsError(false)` at the start of this function, to handle the case of retrying after there was an error

frontend/src/components/TransferCreditHolder.tsx
```diff
@@ -127,10 +127,10 @@ class TransferCreditsComponent extends React.Component<
     if (this.props.transferCredits == null) {
       return null;
     }
-    return this.props.transferCredits.map((scheduleCourse, index) => {
+    return this.props.transferCredits.map((scheduleCourse, _) => {
```
NilsBacke, line 130: We can just remove the _ all together

frontend/src/home/Home.tsx
NilsBacke: Just curious, why is this?

NilsBacke:
This looks really really sweet! Also it's nice that we can use the loading screen for anything now. Is the transition between rendering the loading screen in the redirect screen -> home wrapper seemless?

frontend/src/home/Home.tsx
ArunJ98: when we clear stuff from redux, we break the home component

ArunJ98:
> This looks really really sweet! Also it's nice that we can use the loading screen for anything now. Is the transition between rendering the loading screen in the redirect screen -> home wrapper seemless?
Yea, it's pretty seamless, it's just the text changing and the loading indicator might jump a little bit.

NilsBacke (approved):

PR #232, opened 2020-12-06
fix add plan popper bug

NilsBacke:


NilsBacke:
[AddPlanPopper Errors on creating a plan based on another plan or on a plan that is based of example schedule.](https://trello.com/c/fN5d6Hnc/240-addplanpopper-errors-on-creating-a-plan-based-on-another-plan-or-on-a-plan-that-is-based-of-example-schedule)

ArunJ98 (approved):
nice catch!

PR #233, opened 2020-12-06
handle bad jwt

NilsBacke:


NilsBacke:
[If we can't authenticate user for whatever reason (if they give us a bad JWT token or something), we should show take them back to Welcome Screen.](https://trello.com/c/tjlDQ7na/239-if-we-cant-authenticate-user-for-whatever-reason-if-they-give-us-a-bad-jwt-token-or-something-we-should-show-take-them-back-to-w)

ArunJ98 (approved):

PR #235, opened 2020-12-06
Better search for add class modal

ArunJ98:
MVP of searching for classes. Major next iteration will be to improve UI and add infinite scroll on classes.


frontend/src/api/index.ts
```diff
@@ -45,3 +64,37 @@ export const fetchCourse = async (
...
   }
 };
+
+export const searchCourses = async (
+  searchQuery: string,
+  minIndex: number = 0,
+  maxIndex: number = 10
+): Promise<ScheduleCourse[]> => {
+  const courses: ScheduleCourse[] = [];
+  const proxyurl = "https://cors-anywhere.herokuapp.com/";
```
NilsBacke, line 74: Just curious, why do we need this?

frontend/src/api/index.ts
```diff
@@ -45,3 +64,37 @@ export const fetchCourse = async (
...
+  const response = await fetch(
+    proxyurl +
+      `https://searchneu.com/search?query=${searchQuery}&termId=202130&minIndex=${minIndex}&maxIndex=${maxIndex}`,
```
NilsBacke, line 77: Should we have a helper to get the current term id (based on current day)?

frontend/src/api/index.ts
ArunJ98: Search needs to enable CORS, this is a temporary measure

frontend/src/api/index.ts
ArunJ98: yea definitely - I'm hoping the "idea of a course" service where termIds don't matter will be ready before we need that though :)

frontend/src/utils/schedule-helpers.ts
```diff
@@ -310,3 +310,45 @@ export function getPositionOfYearInSchedule(
...
+    // courseToAdd's classId is an int, so we're casting in order to compare accurately
+    if (
+      String(courseToAdd.classId) === String(course.classId) &&
+      courseToAdd.subject === course.subject
+    ) {
+      return true;
+    }
+  }
+  return false;
+}
```
andrewjleung, line 331: Could we simplify this to avoid needing to explicitly return `true` or `false` based upon a condition?

```typescript
function isCourseInTerm(courseToAdd: ScheduleCourse, term: DNDScheduleTerm) {
    return term.classes.some((course) => String(courseToAdd.classId) === String(course.classId) && courseToAdd.subject === course.subject)
}
```

frontend/src/utils/schedule-helpers.ts
```diff
@@ -310,3 +310,45 @@ export function getPositionOfYearInSchedule(
...
+      isCourseInTerm(courseToAdd, schedule.yearMap[year].spring) ||
+      isCourseInTerm(courseToAdd, schedule.yearMap[year].fall) ||
+      isCourseInTerm(courseToAdd, schedule.yearMap[year].summer1) ||
+      isCourseInTerm(courseToAdd, schedule.yearMap[year].summer2)
+    ) {
+      return true;
+    }
+  }
+  return false;
+}
```
andrewjleung, line 354: Same as the above comment

frontend/src/api/index.ts
```diff
@@ -45,3 +64,37 @@ export const fetchCourse = async (
     return null;
   }
 };
+
+export const searchCourses = async (
+  searchQuery: string,
+  minIndex: number = 0,
+  maxIndex: number = 10
```
andrewjleung, line 71: Just curious what these query params are used for, do they just specify what range of results to return like for pagination?

frontend/src/components/AddClassSearchModal.tsx
```diff
@@ -0,0 +1,228 @@
...
+          <SearchResultsScrollContainer>
+            {isLoading ? (
+              <Loading>
+                <LinearProgress color="secondary" />
+              </Loading>
+            ) : searchedCourseResults.length === 0 ? (
+              <NoResultContainer> No results </NoResultContainer>
+            ) : (
+              searchedCourseResults
+            )}
```
andrewjleung, line 201: For readability's sake since this JSX is getting pretty large, could we split out some of this, especially the conditional renders within `SearchResultsScrollContainer` and `AddedClassesContainer`, into separate components?

frontend/src/api/index.ts
ArunJ98: yup, it's just the range of results search will give back to us

frontend/src/components/AddClassSearchModal.tsx
ArunJ98: definitely, good call

frontend/src/utils/schedule-helpers.ts
ArunJ98: yea that makes sense, this code was copied from the current add class component, probably want to clean this up for sure.

frontend/src/state/index.ts
```diff
@@ -208,4 +208,20 @@ export const getIsAdvisorFromState = (state: AppState) => {
  * SAFE GET FUNCTIONS
  * Should only need to use these where there is not guaranteed to have a user
  */
-export const safelyGetIsAdvisorFromState = (state: AppState) => state.userState.user?.isAdvisor;
\ No newline at end of file
+
+export const safelyGetActivePlanFromState = (state: AppState) => {
```
ArunJ98, line 212: wanted to keep at least this function and safelyGetActivePlanScheduleFromState as we use the add class modal in the onboarding flow. I think we can remove the other ones but also no harm in keeping it. Thoughts?

frontend/src/utils/course-helpers.ts
```diff
@@ -19,3 +19,31 @@ export async function getScheduleCoursesFromSimplifiedCourseDataAPI(
   );
   return converetedCourses;
 }
+
+function isINEUPrereqCourse(val: INEUPrereq): val is INEUPrereqCourse {
```
ArunJ98, line 23: just a type guard

frontend/src/components/AddClassSearchModal.tsx
```diff
@@ -0,0 +1,308 @@
...
+  const [searchQuery, setSearchQuery] = useState("");
+  const [isLoading, setIsLoading] = useState(false);
+  const [searchedCourses, setSearchedCourses] = useState(EMPTY_COURSE_LIST);
+  const [selectedCourses, setSelectedCourses] = useState(EMPTY_COURSE_LIST);
+  const { schedule } = useSelector((state: AppState) => ({
+    schedule: safelyGetActivePlanScheduleFromState(state),
+  }));
+
+  const SearchResult = (props: SearchResultProps) => {
+    let showCourseInSchedyleError = false;
```
NilsBacke, line 164: typo lol

frontend/src/components/AddClassSearchModal.tsx
```diff
@@ -0,0 +1,308 @@
...
+
+export const AddClassSearchModal: React.FC<
+  AddClassSearchModalProps
+> = props => {
+  const [searchQuery, setSearchQuery] = useState("");
+  const [isLoading, setIsLoading] = useState(false);
+  const [searchedCourses, setSearchedCourses] = useState(EMPTY_COURSE_LIST);
+  const [selectedCourses, setSelectedCourses] = useState(EMPTY_COURSE_LIST);
+  const { schedule } = useSelector((state: AppState) => ({
+    schedule: safelyGetActivePlanScheduleFromState(state),
```
NilsBacke, line 160: I don't think we need to do this safely (the only way to access this modal is through the home screen, which will only render if an active plan exists?)

frontend/src/components/AddClassSearchModal.tsx
```diff
@@ -0,0 +1,308 @@
...
+            {props.course.subject + " " + props.course.classId}
+          </SubjectId>
+        </ResultInfoContainer>
+        {showCourseInSchedyleError ? (
+          <Tooltip title="Course already in schedule" aria-label="add">
+            <AddClassError>!</AddClassError>
+          </Tooltip>
+        ) : (
+          <AddClassButton
+            onClick={async () => {
```
NilsBacke, line 184: Can we pull this out into a function in the component as opposed to in the render?

frontend/src/components/AddClassSearchModal.tsx
```diff
@@ -0,0 +1,308 @@
...
+              if (selectedCourses.indexOf(props.course) < 0) {
```
NilsBacke, line 185: Super nit-picky but I think `.includes` would be more clear here

frontend/src/state/index.ts
NilsBacke: I think it's best to keep them. We only want to use the safe ones if we have to.

NilsBacke (approved):

PR #238, opened 2020-12-08
auto save plans

NilsBacke:


frontend/src/home/AutoSavePlan.tsx
```diff
@@ -0,0 +1,89 @@
...
+  );
+
+  useEffect(() => {
+    dispatch(setActivePlanStatusAction("Waiting to Update"));
+    debouncedUpdate(activePlan);
+  }, [
+    JSON.stringify(activePlan.schedule),
+    activePlan.coopCycle,
+    activePlan.major,
+    activePlan.name,
```
NilsBacke, line 54: Only update when values that prompt an update change

frontend/src/home/Home.tsx
```diff
@@ -218,7 +219,9 @@ class HomeComponent extends React.Component<Props> {
   }
 
   componentDidUpdate(nextProps: Props) {
-    if (nextProps.warnings !== this.props.warnings) {
+    if (
+      JSON.stringify(nextProps.warnings) !== JSON.stringify(this.props.warnings)
```
NilsBacke, line 223: This never did work properly lol

NilsBacke:
[Implement some form of auto-saving plans](https://trello.com/c/mQ3DVzTB/149-implement-some-form-of-auto-saving-plans)

frontend/src/home/AutoSavePlan.tsx
```diff
@@ -0,0 +1,89 @@
...
+      await updatePlan(activePlan);
+    }, DEBOUNCE_SAVE_DELAY_MS),
+    []
+  );
+
+  useEffect(() => {
+    dispatch(setActivePlanStatusAction("Waiting to Update"));
+    debouncedUpdate(activePlan);
+  }, [
+    JSON.stringify(activePlan.schedule),
```
NilsBacke, line 51: A little worried that this might become slow because a schedule is a large object (although I did try it with a full example schedule and it seemed fine)

imanmoreira: Does the value need to be a string here for useEffect to work?

frontend/src/home/AutoSavePlan.tsx
```diff
@@ -0,0 +1,101 @@
...
+    activePlanStatus === "Waiting to Update" ||
+    activePlanStatus === "Updating"
+  ) {
+    return (
+      <Container>
+        Saving... <LoopIcon style={{ paddingLeft: 6 }} />
+      </Container>
+    );
+  }
+  if (activePlanStatus === "Up To Date") {
```
imanmoreira, line 92: This could probably just be an else case since the type ActivePlanAutoSaveStatus should only be able to be these three

imanmoreira (changes requested):
Overall code looks really good. I'm getting some buggy behavior though with course warnings, where when you drag it it looks good initially, but then the moment it finishes saving the course specific warnings disappear. I think once that's ironed out it should be all good

frontend/src/home/AutoSavePlan.tsx
NilsBacke: It doesn't have to be a string, but useEffect uses shallow equal, so for objects it's safest to stringify them

NilsBacke:
Course warnings bug fixed!

frontend/src/home/AutoSavePlan.tsx
```diff
@@ -0,0 +1,100 @@
...
+    const token = getAuthToken();
+    dispatch(setActivePlanStatusAction("Updating"));
+    await updatePlanForUser(
+      userId,
+      token,
+      activePlan.id,
+      convertPlanToUpdatePlanData(activePlan)
+    ).then(response => {
+      batch(() => {
+        dispatch(updateActivePlanAction(response.plan));
```
imanmoreira, line 77: I don't think it should update the schedule in the reducer after it finishes updating. At the moment, if you change the schedule while it's still saving, those new changes get overriden

imanmoreira (changes requested):

frontend/src/home/AutoSavePlan.tsx
imanmoreira: Also realizing that the timestamp updates here, so maybe switch it to just update the active plan's timestamp

imanmoreira (approved):

PR #242, opened 2020-12-10
fix race condition in home wrapper component

NilsBacke:


NilsBacke:
[Solve race condition in HomeWrapper when loading all user plans](https://trello.com/c/OVnBeuBZ/248-solve-race-condition-in-homewrapper-when-loading-all-user-plans)

ArunJ98 (approved):

PR #244, opened 2020-12-10
Requesting a plan approval

ArunJ98:
image


NilsBacke (approved):

imanmoreira:
[Create Approve Plan Popper](https://trello.com/c/RYuNuodf/253-create-approve-plan-popper)

.gitignore
```diff
@@ -126,4 +126,5 @@ yarn-error.log*
 !index.html
 
 .idea/
-graduatenu.code-workspace
\ No newline at end of file
+graduatenu.code-workspace
+api/config/local_env.yml
```
NilsBacke, line 130: What's in this file?

api/app/views/notification_mailer/approved_email.html.erb
```diff
@@ -0,0 +1,12 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta content='text/html; charset=UTF-8' http-equiv='Content-Type' />
+  </head>
+  <body>
+    <p>Hi <%= @student.full_name %>,</p>
+    <p><%= @advisor.full_name %> has approved your plan! Woohoo!</p>
+    <p>Plan name:  <%= @plan.name %> <br>
+    Plan ID:  <%= @plan.id %></p>
```
NilsBacke, line 10: Plan ID doesn't mean anything to the student, so we should probably remove this

api/db/schema.rb
```diff
@@ -29,7 +29,11 @@
     t.integer "course_counter"
     t.boolean "is_currently_being_edited", default: false, null: false
     t.datetime "last_viewed"
+<<<<<<< HEAD
+    t.json "approved_schedule"
+=======
```
NilsBacke, line 34: conflict

api/app/controllers/mail_controller.rb
```diff
@@ -0,0 +1,24 @@
+class MailController < ApplicationController
```
NilsBacke, line 1: I think it would make more a lot more sense to have this as part of the `PlansController`. `send_request_approval_email` could be like `request_approval`.

Also do we use `send_plan_approved_email`

frontend/src/home/Home.tsx
```diff
@@ -379,9 +380,9 @@ class HomeComponent extends React.Component<Props> {
 
   async updatePlan(showAlert = true) {
     const token = getAuthToken();
+    console.log(JSON.stringify(this.props.activePlan));
```
NilsBacke, line 383: console log

frontend/src/home/RequestFeedbackPopper.tsx
```diff
@@ -0,0 +1,157 @@
...
+    );
+  };
+
+  const RequestApprovalButton = () => {
+    return (
+      <PrimaryButton
+        disabled={selectedAdvisor === ""}
+        onClick={async () => {
+          // Todo: Trigger backend API call
+          await sendEmail(email, findAdvisorEmail(selectedAdvisor), planId);
```
NilsBacke, line 128: Do we need this TODO anymore?

frontend/src/utils/useDebouncedEffect.js
```diff
@@ -0,0 +1,15 @@
+import { useCallback, useEffect } from "react";
+
+export const useDebouncedEffect = (effect, delay, deps) => {
```
NilsBacke, line 3: This is fire. I think we should copy what office hours does though and put this in a hooks folder. Like `frontend/src/hooks/useDebouncedEffect.js`

api/app/controllers/mail_controller.rb
NilsBacke: Also I think we should add a field onto `Plan` called something like `is_requesting_approval`, so on the advisor side we can show the appropriate alerts/text.

.gitignore
ArunJ98: api key for twilio send grid

api/app/controllers/mail_controller.rb
ArunJ98: Added a field that stores the timestamp of when an approval is requested. We reset the nil when the plan is approved. By using a timestamp instead of a boolean, we can do things like throttle approval requests etc...

api/app/controllers/plans_controller.rb
```diff
@@ -130,6 +167,6 @@ def set_user_plan
   end
 
   def authorized
-    @current_user_id == Integer(params[:user_id])
+    @current_user_id == Integer(params[:user_id]) || @user.is_advisor
```
ArunJ98, line 170: An advisor can pretty much do any action so I think we can just put this here instead of on every instance

api/config/application.rb
```diff
@@ -24,6 +24,13 @@ class Application < Rails::Application
     # Initialize configuration defaults for originally generated Rails version.
     config.load_defaults 6.0
 
+    config.before_configuration do
```
ArunJ98, line 27: we add the environment variables here

frontend/src/home/EditPlanPopper.tsx
```diff
@@ -309,6 +310,24 @@ export class EditPlanPopperComponent extends React.Component<
     );
   }
 
+  renderResetToApprovedButton() {
```
ArunJ98, line 313: idk if we really need this tbh, gonna leave it in for now and might remove it in set primary plan PR

frontend/src/services/AdvisorService.ts
```diff
@@ -1,30 +1,38 @@
...
-  searchQuery: string,
-  pageNumber: number,
-  userToken: string
-) =>
+export const getStudents = (searchQuery: string, pageNumber: number) =>
   fetch(`/api/users/students?search=${searchQuery}&page=${pageNumber}`, {
     method: "GET",
     headers: {
-      Authorization: "Token " + userToken,
+      Authorization: "Token " + getAuthToken(),
```
ArunJ98, line 12: are there any drawbacks to just doing this that I'm missing? Figured we don't need to use getAuthToken every file we call an endpoint

ArunJ98 (commented):
Right now, emails are broken. Not sure why, gonna continue to debug that but wanted to get this merged for advisor workflow

frontend/src/services/AdvisorService.ts
NilsBacke: Don't think so. We might as well refactor to just do this in every API call (in the future)

api/config/application.rb
NilsBacke: We should actually do this in `development.rb` so it only runs in dev mode (locally).

NilsBacke (approved):

PR #245, opened 2020-12-11
Save null (not empty string) in db for major, coopCycle, catalogYear if blank

NilsBacke:
- We should use null instead of undefined/empty string in a lot of places

- Only go to completed courses screen if the user selected a major

- Overhaul onboarding info screen to handle major given from khoury data


NilsBacke:
[Save null (not empty string) in db for major, coopCycle, catalogYear if blank](https://trello.com/c/mqLgfSYU/252-save-null-not-empty-string-in-db-for-major-coopcycle-catalogyear-if-blank)

ArunJ98 (approved):

PR #250, opened 2020-12-13
Abstract schedule

imanmoreira:
Abstracting schedule out into a separate component and creating a non-editable version

Note for whoever is testing: You can test it by going to the profile page after logging in. It's a little buggy since the redux schedules don't always persist correctly when the profiles page refreshes because of the redirect but it should work when you go directly from /home to /profile.


frontend/src/components/SemesterBlock.tsx
```diff
@@ -74,19 +77,71 @@ interface ReduxDispatchSemesterBlockProps {
...
+    <OutsideContainer>
+      <Container warning={false}>
+        <ClassListWrapper>
+          {props.warnings.length > 0 ? (
+            <Tooltip
+              title={renderTooltip(props.warnings)}
+              placement="top"
+              arrow
+            >
+              <>{renderBody(props)}</>
```
NilsBacke, line 133: Do we need the `<>`?

frontend/src/components/Year/Year.tsx
```diff
@@ -48,33 +49,52 @@ class YearComponent extends React.Component<Props> {
...
             fallStatus={schedule.yearMap[year].fall.status}
             springStatus={schedule.yearMap[year].spring.status}
             summer1Status={schedule.yearMap[year].summer1.status}
             summer2Status={schedule.yearMap[year].summer2.status}
+            schedule={schedule}
+            isEditable={isEditable}
           />
         </YearTopWrapper>
-        {isExpanded && (
+        {(!isEditable || isExpanded) && (
```
NilsBacke, line 80: I dont think we need this here as we have `const isExpanded = !this.props.closedYears.has(index) || !isEditable`

frontend/src/components/Year/YearTop.tsx
```diff
@@ -183,11 +171,11 @@ const textContainerStyle: React.CSSProperties = {
...
   alignItems: "center",
   flex: 1,
+  whiteSpace: "nowrap",
+  overflow: "hidden",
 };
 
-const mapStateToProps = (state: AppState) => ({
-  schedule: getActivePlanScheduleFromState(state),
-});
+const mapStateToProps = (state: AppState) => ({});
```
NilsBacke, line 178: Don't need this

frontend/src/home/Home.tsx
```diff
@@ -420,42 +270,60 @@ class HomeComponent extends React.Component<Props> {
...
-                {this.renderTransfer()}
-              </Container>
-            </LeftScroll>
-          </DragDropContext>
-        </OuterContainer>
+        <EditableSchedule sidebarPresent={true} transferCreditPresent={true}>
+          {this.renderPlanHeader()}
+        </EditableSchedule>
       </>
+      // <>
```
NilsBacke, line 277: Can we delete these comments or is it best to save them (we always have git history if we need to revert)?

NilsBacke (approved):

NilsBacke:
[Abstracting Schedule from Home Page](https://trello.com/c/6YLVOVEy/246-abstracting-schedule-from-home-page)

frontend/src/home/Home.tsx
imanmoreira: Nah you're def right, I left this accidently when trying to resolve some merge conflicts

frontend/src/App.tsx
```diff
@@ -26,7 +26,8 @@ export const App = ({ store }: { store: Store }) => {
           <ProtectedRoute path="/home" component={HomeWrapper} />
           <ProtectedRoute path="/redirect" component={RedirectScreen} />
           <ProtectedRoute path="/onboarding" component={OnboardingInfoScreen} />
-          <ProtectedRoute path="/profile" component={Profile} />
+          {/* ****FOR TESTING! DELETE BEFORE MERGING **** */}
```
William-Cunningham, line 29: Do we want to remove this?

frontend/src/components/SemesterBlock.tsx
imanmoreira: Unfortunately yeah, it's because renderBody returns a list and tooltip can only have one child

PR #251, opened 2020-12-17
Trim whiteface on both side of search

ArunJ98:


ArunJ98:
[Bug: In search bar we need to strip user input of white space on each side.](https://trello.com/c/amUYVtJz/52-bug-in-search-bar-we-need-to-strip-user-input-of-white-space-on-each-side)

imanmoreira (approved):

PR #254, opened 2020-12-20
Template page

justineluo:
advising templates page, front-end changes only. Modified from the Figma wireframe to be more consistent with student and notification page design


justineluo:
[Create first Template Page](https://trello.com/c/WEF2fQFK/198-create-first-template-page)

api/test/fixtures/users.yml
```diff
@@ -18,6 +18,7 @@
 #  remember_created_at    :datetime
 #  reset_password_sent_at :datetime
 #  reset_password_token   :string
+#  username               :string
```
NilsBacke, line 21: Ty running `cd api && docker-compose run server rails db:migrate` and see if that fixes this bc this username field should not be here

frontend/src/advising/Templates.tsx
```diff
@@ -1,10 +1,295 @@
...
+    //     setTemplates(currentTemplates.concat(TemplatesAPI.templates));
+    //     setPageNumber(TemplatesAPI.nextPage);
+    //     setIsLastPage(TemplatesAPI.lastPage);
+    //     setIsLoading(false);
+    //   })
+    //   .catch(err => console.log(err));
+  };
+
+  useEffect(() => {
+    setTemplates(EMPTY_TEMPLATES_LIST);
```
NilsBacke, line 190: I don't think we need this here, we have the initial state value

frontend/src/advising/Templates.tsx
```diff
@@ -1,10 +1,295 @@
...
+    fetchTemplates(EMPTY_TEMPLATES_LIST, 0);
+  }, [props.searchQuery, token]);
+
+  return (
+    <TemplatesContainer>
+      {isLoading ? (
```
NilsBacke, line 196: We can just do `isLoading && ...`

frontend/src/services/AdvisorService.ts
```diff
@@ -14,3 +14,20 @@ export const getStudents = (
...
+ * @param searchQuery  the search query
+ * @param pageNumber  page number for the query
+ * @param userToken the JWT token of the user
+ */
+export const getTemplates = (
+  searchQuery: string,
+  pageNumber: number,
+  userToken: string
+) =>
+  fetch(`/api/users/students?search=${searchQuery}&page=${pageNumber}`, {
```
NilsBacke, line 28: Is this just temporary?

andrewjleung: Yeah, we just had it in as a placeholder since the API isn't set up yet, but we'll remove it for now to avoid confusion.

NilsBacke (approved):

PR #255, opened 2020-12-21
set correct year on each semester

NilsBacke:


imanmoreira (approved):
LOL good catch

frontend/src/components/common/FullPageLoading.tsx
```diff
@@ -2,6 +2,7 @@ import React from "react";
 import styled from "styled-components";
```
NilsBacke, line 2: Somehow this file got reverted to about 2 weeks ago. I bring the changes back here

NilsBacke:
[Can't delete a class with a name that is too long](https://trello.com/c/kou2wr2t/260-cant-delete-a-class-with-a-name-that-is-too-long)

PR #259, opened 2020-12-24
expanded student plan

NilsBacke:


NilsBacke:
[Create expanded student plan view (Both editable view and non editable view)](https://trello.com/c/7KdMgZRl/256-create-expanded-student-plan-view-both-editable-view-and-non-editable-view)

frontend/src/advising/ManageStudents.tsx
```diff
@@ -282,62 +337,139 @@ const StudentPreview = ({ id }: StudentPreviewProps) => {
       dispatch(resetUserAction());
     };
   }, []);
+
+  if (viewMode === "overview") {
```
imanmoreira, line 341: So just based on the figma, the expanded schedule views didn't need to have the search bar visible above it, so that part still could be a separate page like you had mentioned before with the student id and plan id. I think it's still fine like this but it might be nice to pull that part out into a separate page since it can be

frontend/src/advising/ManageStudents.tsx
```diff
@@ -282,62 +337,139 @@ const StudentPreview = ({ id }: StudentPreviewProps) => {
...
-          <>
-            <PlanTitle>
-              <TitleText>{planName}</TitleText>
-            </PlanTitle>
-            <ButtonHeader>
-              <IconButton>
-                <Edit />
+    <FullScheduleViewContainer>
+      <BackToStudentLink onClick={props.onFullScreenPress}>
+        Back to {fullName}
```
imanmoreira, line 434: I think this would look cleaner if this was more like a back arrow icon and in line with the student information currently being displayed below it. It feels a little awkward right now

image

imanmoreira (changes requested):
Hey! Only change that I feel it really necessary is switching to the back arrow icon. 
I added a comment about making this a separate page but I'd be okay to merge without changing that.

frontend/src/advising/ManageStudents.tsx
NilsBacke: Yeah I did notice this, and I decided to just leave the search bar in because it doesn't hurt. Once we move the Single Student View and this page to their own separate routes, we can do what we feel is best with the search bar on those pages.

imanmoreira (approved):

PR #265, opened 2020-12-26
Templates backend

NilsBacke:


api/Gemfile.lock
```diff
@@ -1,71 +1,71 @@
 GEM
   remote: https://rubygems.org/
   specs:
-    actioncable (6.0.2.2)
-      actionpack (= 6.0.2.2)
+    actioncable (6.0.3.4)
```
NilsBacke, line 4: Updated all Rails gems

NilsBacke:
Probably best to review this PR commit by commit. Each commit is labelled appropriately.

NilsBacke:
[Hook up backend to the first templates page.](https://trello.com/c/ZMrRMfQ2/263-hook-up-backend-to-the-first-templates-page)

NilsBacke:
Took out the redux commit, turns out we won't need it once we transition to each page being on its own route

api/test/controllers/templates_controller_test.rb
```diff
@@ -0,0 +1,168 @@
+require 'test_helper'
```
ArunJ98, line 1: nice tests 💯

frontend/src/services/TemplateService.ts
```diff
@@ -0,0 +1,57 @@
...
+  pageNumber: number,
+  userId: number,
+  userToken: string
+): Promise<TemplatesAPI> =>
+  fetch(
+    `/api/users/${userId}/templates?search=${searchQuery}&page=${pageNumber}`,
+    {
+      method: "GET",
+      headers: {
+        Authorization: "Token " + userToken,
```
ArunJ98, line 26: can we replace userToken here with getAuthToken()

frontend/src/services/TemplateService.ts
```diff
@@ -0,0 +1,57 @@
...
+export const createTemplate = (
+  userId: number,
+  userToken: string,
+  template: ICreateTemplatePlan
+): Promise<IFolderData> =>
+  fetch(`/api/users/${userId}/templates`, {
+    method: "POST",
+    body: JSON.stringify({ templatePlan: template }),
+    headers: {
+      Authorization: "Token " + userToken,
```
ArunJ98, line 55: same as above

ArunJ98 (commented):
sweeeeet! this is awesome, just left some minor comments/clarifications

api/app/controllers/templates_controller.rb
```diff
@@ -0,0 +1,93 @@
...
+      if @user == nil
+        render json: {error: "User not found."}, status: 404
+      end
+    else
+      render json: {error: "Unauthorized."}, status: :unprocessable_entity
+    end
+  end
+
+  #sets the user whose templates are being searched for (in most cases, their own)
+  def set_searched_advisor
```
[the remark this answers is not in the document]
NilsBacke: So for now they're going to be the same, but my thinking is it's best to separate them. `@searched_advisor` is going to be the advisor whose templates we're editing/creating/deleting. `@user` is the user making the request. Right now in the app, they're always going to be the same, but in the future, we may want to allow any advisor to edit any other advisor's plans (shared plans).

ArunJ98 (approved):

PR #267, opened 2020-12-27
put manage students flow all in separate routes

NilsBacke:


NilsBacke:
[Move Single Student View and Expanded Student View to separate routes with id's in params](https://trello.com/c/mXlq0PwH/270-move-single-student-view-and-expanded-student-view-to-separate-routes-with-ids-in-params)

frontend/src/advising/ManageStudents/StudentView.tsx
```diff
@@ -0,0 +1,226 @@
...
+      setFetchingStudent(false);
+    });
+    findAllPlansForUser(id).then((plans: IPlanData[]) => {
+      dispatch(setUserPlansAction(plans, 2020));
+      if (!plans || !plans.length || !plans[0].schedule) setNoPlans(true);
+    });
+  }, []);
+
+  return (
+    <Container>
```
ArunJ98, line 145: nit but can we break up this JSX? this render function is pretty big, could we split it into smaller components? For example, we could have the StudentPreviewContainer and SchedulePreviewContainer as different components

frontend/src/services/AdvisorService.ts
```diff
@@ -1,5 +1,18 @@
 import { getAuthToken } from "../utils/auth-helpers";
 
+export interface IAbrStudent {
+  fullName: string;
```
ArunJ98, line 4: I added a IAbrUser that's very similar to this to the types folder. Can we merge the two?

ArunJ98:
Overall looks really good, just some minor comments

frontend/src/services/AdvisorService.ts
NilsBacke: Seems like they're for two different use cases, so I renamed the existing one to `IAbrAdvisor`

ArunJ98 (approved):

PR #275, opened 2020-12-30
Allow advisors to approve plans

ArunJ98:


frontend/src/components/common/SnackbarAlert.tsx
```diff
@@ -0,0 +1,50 @@
+import React from "react";
```
NilsBacke, line 1: I believe we use Toasts for warnings on the student side. Is this something different or could we maybe use the same Toasts?

NilsBacke:
[Add approve button to expanded student plan view and hook up to backend](https://trello.com/c/ziPgEWte/272-add-approve-button-to-expanded-student-plan-view-and-hook-up-to-backend)

frontend/src/components/common/SnackbarAlert.tsx
imanmoreira: I believe it's the same functionality but I think it would be nice if we don't have to add in all of the useToast/toastProvider stuff that was necessary on the student end. Also if I recall correctly we have a bit of a mix between snackbar and toast being used on the students end (undo button and delete plan is snackbar and warnings are Toast)

imanmoreira (approved):

PR #277, opened 2021-01-01
Set primary plan

ArunJ98:
Allows students to set a primary plan for an advisor. We should iterate on what the best UI is to indicate a primary plan to a user. Maybe we replace "Plan of Study" with "Your Primary Plan" or the plan name if it's not primary?

### DIFF
--- a/api/app/controllers/plans_controller.rb
+++ b/api/app/controllers/plans_controller.rb
@@ -116,6 +116,20 @@ class PlansController < ApplicationController
     end
   end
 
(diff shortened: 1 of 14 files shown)


ArunJ98:
[Allow students to mark one plan as their primary plan.](https://trello.com/c/Gi76ad1E/161-allow-students-to-mark-one-plan-as-their-primary-plan)

NilsBacke (approved):

PR #280, opened 2021-01-03
Create separate route for Templates

imanmoreira:
- separated the routes for the templates page
- Making different colored buttons common components


frontend/src/advising/Templates/TemplateListPage.tsx
```diff
@@ -188,7 +187,7 @@ const TemplatesList = ({ searchQuery }: TemplatesListProps) => {
   }, [searchQuery]);
 
   return (
-    <TemplatesContainer>
+    <>
```
imanmoreira, line 190: This is to get rid of the weird double red outline display that was happening on the TemplateListPage

NilsBacke:
[Move all templates pages to be their own route (templates list, new template)](https://trello.com/c/ZwEvRNrY/277-move-all-templates-pages-to-be-their-own-route-templates-list-new-template)

frontend/src/home/Home.tsx
```diff
@@ -132,16 +133,6 @@ const LogoutButton = styled.div`
   margin-right: 8px !important;
 `;
 
-const ColorButton = withStyles((theme: Theme) => ({
```
NilsBacke, line 135: I believe @andrewjleung was also working on similar button refactors. What we have here looks good, just @'ing to make him aware.

NilsBacke (approved):

frontend/src/home/Home.tsx
andrewjleung: @justineluo and I ended up leaving out button refactors, so this should be all fine 👍

andrewjleung (approved):

PR #281, opened 2021-01-06
Log changes an advisor makes to a schedule and UI to show changes

ArunJ98:
Still need to (in order of importance): Don't think all of these need to happen in this PR 
- [x]  Automatically scroll down to bottom of comments/changelog
- [x]  Log student changes
- [x]  Add comment/changelog to students
- [ ]  Add "Highlight" tab that shows most recent changes and comments in one feed

image


ArunJ98:
[Auto add all advisor schedule changes to the comments box](https://trello.com/c/0GzoEQcx/271-auto-add-all-advisor-schedule-changes-to-the-comments-box)

api/db/migrate/20210102222850_create_plan_changelogs.rb
```diff
@@ -0,0 +1,11 @@
+class CreatePlanChangelogs < ActiveRecord::Migration[6.0]
+  def change
+    create_table :plan_changelogs do |t|
+      t.string :log, null: false
+      t.bigint :author_id, null: false
```
ArunJ98, line 5: author_id is needed so we can use this to determine if the current user is the author of the log or not without relying on the name (can be duplicate)

frontend/src/services/PlanService.ts
```diff
@@ -150,3 +152,93 @@ export const setPrimaryPlan = (userId: number, planId: number) =>
...
+      Authorization: "Token " + getAuthToken(),
+    },
+  }).then(response => response.json());
+
+/**
+ * Gets comments for a plan
+ * @param planId The id of the plan
+ * @param studentId The id of the student
+ */
+export const fetchComments = (planId: number, studentId: number) =>
```
ArunJ98, line 201: I think comments belong in the plan service as students will also be viewing/adding comments

frontend/src/utils/ScheduleChangeTracker.ts
```diff
@@ -0,0 +1,71 @@
+import { convertTermIdToSeasonString, convertTermIdToYear } from ".";
```
ArunJ98, line 1: This is the core of batching the changes

api/app/models/plan_changelog.rb
```diff
@@ -0,0 +1,19 @@
...
+#  updated_at :datetime         not null
+#  author_id  :bigint(8)        not null
+#  plan_id    :bigint(8)
+#
+# Indexes
+#
+#  index_plan_changelogs_on_plan_id  (plan_id)
+#
+class PlanChangelog < ApplicationRecord
+    belongs_to :plan
```
NilsBacke, line 18: We can do `belongs_to :author, class_name: 'User', foreign_key: :author_id` here to be able to do `plan_changelog.author`

api/db/migrate/20210102222850_create_plan_changelogs.rb
```diff
@@ -0,0 +1,11 @@
...
+      t.string :author, null: false
```
NilsBacke, line 6: We don't need this as the author name can be retrieved through the author id. (and with my comment above, we can just do `planChangelog.author.name`

api/test/factories/plan_changelogs.rb
```diff
@@ -0,0 +1,21 @@
+# == Schema Information
```
NilsBacke, line 1: You can delete this file

frontend/src/Onboarding/TransferableCreditScreen.tsx
```diff
@@ -218,7 +218,7 @@ const TransferableCreditScreen: React.FC = () => {
         );
       }
 
-      createPlanForUser(userId!, token, {
+      createPlanForUser(userId!, {
```
NilsBacke, line 221: Can we also remove where we define token in this file? (Unless we use it elsewhere)

frontend/src/home/Home.tsx
```diff
@@ -215,9 +217,12 @@ class HomeComponent extends React.Component<Props> {
       // switched plan
       this.callUpdatePlanLastViewedOnInterval();
     }
+    window.addEventListener("beforeunload", this.sendPlanUpdates);
   }
 
   componentWillUnmount() {
+    this.sendPlanUpdates();
+    window.removeEventListener("beforeunload", this.sendPlanUpdates);
```
NilsBacke, line 225: We may want to keep an eye on this, I've heard people say that making an API call on `beforeunload` or `unload` is unreliable, but I think this is fine until we determine it's not.

frontend/src/utils/ScheduleChangeTracker.ts
```diff
@@ -0,0 +1,71 @@
...
+
+/**
+ * The ScheduleChangeTracker class defines the `getInstance` method that lets clients access
+ * the unique ScheduleChangeTracker instance.
+ */
+class ScheduleChangeTracker {
```
NilsBacke, line 7: I think this is ok for the time being. Other solutions I can think of are to create a new redux slice, or make use of React contexts. I think either of those would be "better practice", but since you already have this and it's working _and we're on a time crunch_ I don't see why we can go for it. 

One thing that we should test is that everything works as expected if an advisor goes into a student, makes some changes, then goes into a different student and makes changes.

frontend/src/components/Schedule/Comments.tsx
```diff
@@ -76,46 +116,239 @@ const CommentHeaderWithDropDown = styled.div`
...
+                <>
+                  <CommentContainer />
+                  <ChangeLogsContainer />
+                  <ContentEnd ref={contentEndRef}></ContentEnd>
+                </>
+              )}
+            </div>
+          )}
+        </ContentContainer>
+        <CommentInput />
```
imanmoreira, line 335: It feels a little weird/unintuitive that you can add comments when on the change log tab since it feels like it's indicating you can comment on the change log. I'd say move it to just comments unless you have some other reason for keeping it for both

imanmoreira (commented):
Probably will have some more later but I figured I'd post these comments for now

frontend/src/components/Schedule/Comments.tsx
```diff
@@ -76,46 +116,239 @@ const CommentHeaderWithDropDown = styled.div`
...
+        {/* Show alert only if comments are not visible */}
```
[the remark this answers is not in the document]
ArunJ98: My thinking was that if comments are visible, they see the comment get added to the view so I felt that another snackbar alert would be too much. Yea I noticed the switching thing too but I felt that while using it, user's wouldn't switch around too much.

frontend/src/components/Schedule/Comments.tsx
ArunJ98: Yea that makes sense, I wanted to make commenting at any point super easy. Maybe we take users back to the comment tab if they comment?

frontend/src/Onboarding/TransferableCreditScreen.tsx
ArunJ98: Yea it's used in another call. I'll put up a separate PR after this one to clean it all up

imanmoreira (approved):
Lgtm

PR #282, opened 2021-01-06
Updating Major Data Request

imanmoreira:
Updating endpoint call for searchneu's staging environment


common/constants.ts
```diff
@@ -0,0 +1,95 @@
+// majors on which the scraper has been verified to run correctly:
+export const supported2018_2019 = [
```
imanmoreira, line 2: Moved these from scraper tests to common so they can be used for the mapping of supported majors

imanmoreira:
[Update graphQL call](https://trello.com/c/8agCY5kW/281-update-graphql-call)

frontend/src/utils/fetchMajorsAndPlans.ts
```diff
@@ -74,8 +73,12 @@ export function fetchMajorsAndPlans() {
           if (res.error) {
             throw res.error;
           }
+          console.log("reached done");
```
ArunJ98, line 76: do we need these console logs?

frontend/src/utils/plan-helpers.ts
```diff
@@ -14,6 +14,7 @@ export function findMajorFromName(
   if (!major) {
     return undefined;
   }
+  console.log(major.yearVersion);
```
ArunJ98, line 17: same as above

frontend/src/utils/fetchMajorsAndPlans.ts
imanmoreira: LOL thanks

ArunJ98 (approved):

PR #291, opened 2021-01-09
concentration backend

justineluo:
We added concentration column to template_plans, plans, users and connected with frontend


frontend/src/models/types.ts
```diff
@@ -264,6 +266,7 @@ export interface IUpdateUserData {
   academic_year?: number;
   graduation_year?: number;
   coop_cycle?: string | null;
+  concentration: string | null;
```
imanmoreira, line 269: this can be optional (?)

imanmoreira (changes requested):
:meow-party:

frontend/src/profile/Profile.tsx
```diff
@@ -189,6 +186,7 @@ const ProfileComponent: React.FC = () => {
             value={val}
             onChange={(event: React.SyntheticEvent<{}>, value: any) => {
               setMajor(value);
+              setCoopCycle("");
```
NilsBacke, line 189: I think we should be setting it to null here

NilsBacke: Yeah we should refactor this screen slightly to use nulls instead of empty strings. The main motivation for this is instead of saving a blank string in the db, we save null which is more reflective of "this user has no ___ set". I think we do this in AddPlanPopper and/or EditPlanPopper

frontend/src/profile/Profile.tsx
```diff
@@ -301,15 +299,10 @@ const ProfileAdvisor = (props: any) => {
...
+      dispatch(setStudentMajorAction(major));
       dispatch(setStudentCatalogYearAction(catalogYear));
       dispatch(setStudentConcentrationAction(concentration));
-      if (coopCycle !== "None Selected") {
-        dispatch(setStudentCoopCycleAction(""));
-      } else {
-        dispatch(setStudentCoopCycleAction(coopCycle || ""));
-      }
-
+      dispatch(setStudentCoopCycleAction(coopCycle));
```
andrewjleung, line 305: I've tested that the dropdowns which display "None selected" dispatch `null` rather than "None selected" even with this change.

frontend/src/profile/Profile.tsx
```diff
@@ -200,7 +198,7 @@ const ProfileComponent: React.FC = () => {
   };
 
   const ProfileCatalogYear = () => {
-    const val = catalogYear != undefined ? String(catalogYear) : "";
+    const val = String(catalogYear || "None selected");
```
andrewjleung, line 201: Just to keep the empty display consistent with the other dropdowns on this page

imanmoreira (approved):

PR #297, opened 2021-01-24
Assign Templates

imanmoreira:
Should allow advisors to
- Assign a student a specific template from the template builder
- Assign a template to the student from the manage single students page
- Apply searches for either of the two
- Delete a plan on assign for either of the two
- Template plans should generate the correct years when it is assigned


frontend/src/advising/ManageStudents/AssignTemplateToUserModal.tsx
```diff
@@ -0,0 +1,81 @@
...
+    setIsLoading: (loading: boolean) => void,
+    setResponse: (response: any, newList: IFolderData[]) => void
+  ) => {
+    setIsLoading(true);
+    getTemplates(userId, searchQuery, page)
+      .then((response: TemplatesAPI) => {
+        setResponse(response, currentList.concat(response.templates));
+        setIsLoading(false);
+      })
+      .catch((err: any) => console.log(err));
```
andrewjleung, line 47: Should we do anything here for error handling beyond logging to console?

andrewjleung (approved):
Really clean modal abstraction! Just one question about error handling but otherwise looks good to me, nice work 👍

frontend/src/advising/ManageStudents/AssignTemplateToUserModal.tsx
imanmoreira: Hmm yeah probably. I think overall we need a more generic way to handle errors because right now it's pretty mismatched. I'll talk to Alex about it and see if she can come up with a nice design/plan for it

PR #299, opened 2021-01-28
WIP: fix templates schedule bug

NilsBacke:
Template Plan's schedules were always null.


NilsBacke:
[Fix templates schedule being null](https://trello.com/c/auJgfaJa/293-fix-templates-schedule-being-null)

PR #300, opened 2021-01-29
sidebar credit tracker

William-Cunningham:


frontend/src/components/Sidebar/Sidebar.tsx
```diff
@@ -31,7 +33,16 @@ const MajorTitle = styled.p`
...
+`;
+
+const CreditTitle = styled.p<any>`
+  font-size: 14px;
+  font-weight: 600;
+  line-height: 16px;
+  margin-right: 12px;
+  margin-left: 4px;
+  margin-bottom: 6px;
+  color: ${props => (props.isGreen ? "green " : "red")};
```
imanmoreira, line 45: Looked at it with the generic colors and they're def a bit jarring. the red we use should be in a Colors file as Northeastern Red and the green being used is `rgba(21,116,62,0.68)`
(remind me to show you how to find out what color is being used for something on the page in the future so you can find it later )

imanmoreira (commented):
Looks great, just a small change on the colors like we talked about before

imanmoreira (approved):

PR #303, opened 2021-01-31
Filter by template name on frontend

ArunJ98:


imanmoreira (changes requested):
There's also a similar template search being done on one of the assign plan modals. Could you add a similar change there?

imanmoreira (approved):

PR #304, opened 2021-02-03
Add option to create from existing or new folder on TemplateInfoPage

andrewjleung:
## Why

We need a way for advisors to create new folders.

## What

- Replace the existing existing folder selection on the `TemplateInfoPage` with a new `FolderSelection` component that provides the option to either choose an existing folder or create a new folder

## Notes

- For the future, I'm going to also revamp the error handling on this page to match that of the other pages, i.e. having individual errors for each form that display when pressing next
- There is some minor styling wonkiness when switching between inputs (they have different heights)
- We are definitely going to need to think about providing the option to delete folders as well + what the intricacies of that are going to be
- We eventually will want to give the option to create an empty folder from the list page

### DIFF
--- a/frontend/src/advising/Templates/TemplateInfoPage.tsx
+++ b/frontend/src/advising/Templates/TemplateInfoPage.tsx
@@ -30,8 +30,6 @@ import {
   planToString,
 } from "../../utils";
 
-const FOLDER_NAME_FIELD_LABEL = "New folder name";
-
 const Container = styled.div`
   margin-left: 30px;
   margin-right: 30px;
@@ -57,6 +55,14 @@ const ButtonContainer = styled.div`
 `;
 
 const InputContainer = styled.div`
+  width: 326px;
+`;
+
+const FolderSelectionContainer = styled.div`
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  gap: 25px;
   width: 326px;
 `;
 
@@ -116,7 +122,7 @@ const DisplayedInputField: React.FC<{ creatingNewFolder: boolean }> = ({
       <NameField
         name={newFolderName}
         setName={setNewFolderName}
-        label={FOLDER_NAME_FIELD_LABEL}
+        label={"New folder name"}
       />
     );
   }
@@ -132,7 +138,7 @@ const DisplayedInputField: React.FC<{ creatingNewFolder: boolean }> = ({
           <TextField
             {...params}
             variant="outlined"
-            label={"Save Plan in Folder"}
+            label={"Existing folder"}
             fullWidth
           />
         )}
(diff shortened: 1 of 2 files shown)


frontend/src/advising/Templates/TemplateInfoPage.tsx
```diff
@@ -67,9 +83,150 @@ interface DropdownProps {
 
 interface NameFieldProps {
   readonly name: string;
-  readonly setTemplateName: (name: string) => void;
+  readonly setName: (name: string) => void;
+  readonly label: string;
+  readonly error: boolean;
```
andrewjleung, line 88: Made this component a bit more abstract for reuse with folder names

frontend/src/advising/Templates/TemplateInfoPage.tsx
```diff
@@ -67,9 +83,150 @@ interface DropdownProps {
...
 
+interface FolderContext {
+  readonly folders: IFolderData[];
+  readonly setSelectedFolderId: (selectedFolderId: number | null) => void;
+  readonly newFolderName: string;
+  readonly setNewFolderName: (newFolderName: string) => void;
+  readonly hasDuplicateFolderName: boolean;
+}
+
+const folderContext = createContext<Partial<FolderContext>>({});
```
andrewjleung, line 99: Context for sharing state from the `NewTemplatesPage` component with its children (so we don't have to keep passing props down over and over)

frontend/src/advising/Templates/TemplateInfoPage.tsx
```diff
@@ -67,9 +83,150 @@ interface DropdownProps {
...
+      value={name}
+      onChange={event => setName(event.target.value)}
+      placeholder=""
+      style={{ width: "100%" }}
+      error={error}
+    />
+  );
+};
+
+const DisplayedFolderInputField: React.FC<{ creatingNewFolder: boolean }> = ({
```
andrewjleung, line 121: This represents solely the conditionally displayed input field, whether that be a dropdown of existing folders or a text field for entering a new folder

frontend/src/advising/Templates/TemplateInfoPage.tsx
```diff
@@ -67,9 +83,150 @@ interface DropdownProps {
...
+        )}
+        onChange={(event, newValue: any) =>
+          setSelectedFolderId(newValue.id || null)
+        }
+      />
+    </FormControl>
+  );
+};
+
+const FolderSelection: React.FC<{}> = () => {
```
andrewjleung, line 181: The entire folder selection component including the picker between both options and the input itself

frontend/src/advising/Templates/TemplateInfoPage.tsx
```diff
@@ -67,9 +83,150 @@ interface DropdownProps {
...
+  if (folders.length < 1) {
+    return (
+      <NameField
+        name={newFolderName}
+        setName={setNewFolderName}
+        label={"New folder name"}
+        error={hasDuplicateFolderName}
+      />
+    );
+  }
```
andrewjleung, line 207: Display only the name field with nothing else if there are no folder, effectively blocking off the "select from existing folders" option if there simply are none to choose from

frontend/src/advising/Templates/TemplateInfoPage.tsx
```diff
@@ -140,115 +314,103 @@ export const NewTemplatesPage: React.FC<RouteComponentProps<{}>> = ({
       catalog_year: catalogYear ? Number(catalogYear) : null,
       folder_id: selectedFolderId,
       folder_name:
-        folders.find(folder => selectedFolderId === folder.id)?.name || null,
+        folders.find(folder => selectedFolderId === folder.id)?.name ||
+        newFolderName ||
+        null,
```
andrewjleung, line 319: Use the new folder name if there is no existing folder selected

imanmoreira (approved):

PR #308, opened 2021-02-07
sort validation based on group type

justineluo:
This is a subtask on this ticket: https://trello.com/c/rkcjckMU/30-update-requirement-group-validation-in-the-sidebar

Sort the requirement groups based on type of IMajorRequirementGroup, from most to least constrained, with most being AND and least being RANGE. 

Example use case for sorting is that we'd want to validation concentration courses before validating electives

### DIFF
--- a/frontend/src/utils/generate-warnings.ts
+++ b/frontend/src/utils/generate-warnings.ts
@@ -249,9 +249,6 @@ export function produceRequirementGroupWarning(
   );
 
   for (const requirementGroup of sortedRequirements) {
-    // todo: if req group is RangeSection or contains a ICourseRange process after all other requirement groups.
-    // this is because these reqs can be satisfied by a wide range of courses, and you don't want it using up
-    // a course that is need to satisfy a more constrained requirement.
     let unsatisfiedRequirement:
       | IRequirementGroupWarning
       | undefined = produceUnsatifiedRequirement(
--- a/frontend/src/utils/generate-warnings.ts
+++ b/frontend/src/utils/generate-warnings.ts
@@ -30,6 +30,7 @@ import {
   Season,
   StatusEnum,
 } from "../../../common/types";
+import { sortOnValues } from "./requirementGroupUtils";
 /*
 CreditRange interface to track the min and max credits for a particular season.
 seasonMax = a number representing the max numebr of credits you can take without over-loading.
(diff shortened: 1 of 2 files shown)


frontend/src/utils/requirementGroupUtils.ts
```diff
@@ -0,0 +1,27 @@
+import { IMajorRequirementGroup } from "../../../common/types";
+
+const reqGroupSortOrder = ["AND", "OR", "RANGE"];
+
+/**
+ *
```
andrewjleung, line 6: Docstring here

frontend/src/utils/requirementGroupUtils.ts
```diff
@@ -0,0 +1,27 @@
...
+ * @param groupMap Mapping of section name to IMajorRequirementGroup validation group
+ */
+export const sortOnValues = (
```
andrewjleung, line 9: We should rename to be more specific about its function:
```suggestion
export const sortRequirementGroupsByConstraint = (
```
Or something else that's more concise, naming is hard 😄

frontend/src/utils/generate-warnings.ts
```diff
@@ -241,11 +242,13 @@ export function produceRequirementGroupWarning(
...
-      major.requirementGroupMap[name];
+
+  const sortedRequirements = sortOnValues(
+    Object.values(major.requirementGroupMap)
+  );
+
+  for (const requirementGroup of sortedRequirements) {
     // todo: if req group is RangeSection or contains a ICourseRange process after all other requirement groups.
     // this is because these reqs can be satisfied by a wide range of courses, and you don't want it using up
     // a course that is need to satisfy a more constrained requirement.
```
andrewjleung, line 254: I believe the sorting within this PR is what this comment is referring to, if so we can remove it

frontend/src/utils/requirementGroupUtils.ts
```diff
@@ -0,0 +1,27 @@
...
+        reqGroupSortOrder.indexOf(first.type) -
+        reqGroupSortOrder.indexOf(second.type);
+      if (diff < 0) {
+        return -1;
+      }
+      if (diff > 0) {
+        return 1;
+      } else {
+        return 0;
+      }
```
andrewjleung, line 25: Here's a concise way to restrict a value within a range like you're doing here!
 
```suggestion
      Math.min(1, Math.max(-1, diff))
```

andrewjleung (changes requested):
Just a few small changes, then LGTM

andrewjleung (approved):

PR #315, opened 2021-02-14
Fix 100% width for folder name on TemplateInfoPage

andrewjleung:
## Why 

There is an issue with how the folder name input field looks if the advisor has no existing folders. This was because the field's width property was set to 100%.

## What
Sets a static width of 326px for the input field to match the other fields.


imanmoreira (approved):
:)

PR #317, opened 2021-02-18
Edit Plan Popper Refactor and Add Ability to Rename Plans

ArunJ98:
WIP, some cleanup still left to do


frontend/src/state/reducers/userPlansReducer.ts
```diff
@@ -183,38 +190,38 @@ export const userPlansReducer = (
...
-          clearSchedule(newSchedule),
-          academicYear,
-          graduationYear
-        );
-        draft.plans[draft.activePlan!].courseCounter = 0;
-
-        // clear all warnings
-        draft.plans[draft.activePlan!].warnings = [];
-        draft.plans[draft.activePlan!].courseWarnings = [];
+        // if (!allPlans) {
```
NilsBacke, line 193: Can we delete this code?

andrewjleung:
Currently there's no validation here for duplicate plan names, IIRC plans are actually uniquely identified by their names, we are going to need to add that to prevent any collisions.

andrewjleung (changes requested):
See above comment

PR #320, opened 2021-03-05
Add additional co-op cycle options 

andrewjleung:
[Ticket](https://trello.com/c/5ST3Q0EJ)

## Why

We want users to be able to select co-op cycles that don't have a corresponding plan.

## What
- Compute and export a static constant which contains co-op cycle options
- Use these options in co-op cycle dropdowns
- In the `EditPlanPopper`, conditionally display the `SET EXAMPLE SCHEDULE` button based upon if there is an available schedule

## Notes
- This only exposes the new co-op cycles and allows them to be selected and persisted, schedules are not yet able to update to match them (I was told this would be future work)
- I did not add this to the `AddPlanPopper` because of the above point


ArunJ98 (approved):

ArunJ98:
Side note: This will probably have some merge conflicts with the EditPlanPopper

andrewjleung:
@ArunJ98 Shouldn't be too bad, just have to update the options within the co-op cycle dropdown in the `EditPlanPopper` to use the `BASE_FORMATTED_COOP_CYCLES` constant 👍

PR #332, opened 2021-04-01
Add advising page errors

TalUsvyatsky:
Add error page for when students don't load with error title, error message, and doggo


justineluo:
[Error Handling redesigns on advisor end](https://trello.com/c/Qi7sgMTv/320-error-handling-redesigns-on-advisor-end)

PR #333, opened 2021-04-01
Add "Upload Plan of Study" Functionality

andrewjleung:
https://trello.com/c/Wm5oApip

## Why

We want advisors to be able to import templates through excel workbooks in the same way they can upload schedules. 

## What

- Added a method `excelToScheduleMultipleSheets` to the `excelParser` for parsing multiple schedules and their names from an excel workbook
- Added a new `ExcelWorkbookUpload` component with the functionality to import and parse a workbook of schedules
- Added a new `PlanUploadPopper` component to the `TemplateListPage` which displays a form for uploading templates
- Pulled out and abstracted folder selection into a separate file for reuse in the new popper
- Pulled out common functionality for using the templates API into a hook for reuse across both the `TemplateListPage` and the `PlanUploadPopper`, use a context to pass relevant pieces
- Added a new endpoint for creating a folder for use when persisting multiple templates into a new folder at once, this operation requires the folder to be created first so that the ID is known when persisting the templates

## Notes

Right now error handling across all the forms is all over the place and needs further consolidation to a single pattern. After looking at it for way too long I think we're better off using a library like [`react-hook-form`](https://react-hook-form.com/) for this task, since things get fairly complex pretty quickly. I'll be looking at this afterwards.

pain


imanmoreira (commented):
Can't find exactly where in the code it is but for the uploading excel spreadsheets, I don't think it should be mandatory to select a major, since it could be several majors

imanmoreira (changes requested):
Other than the small change about majors, this looks amazing and it works great!

andrewjleung:
@imanmoreira I initially didn't include major but it was rejected by the templates API since it was required, all the fields that are on the popper are there because the API required them

imanmoreira (approved):

PR #337, opened 2021-04-06
Hook up Plan Appointments with Backend

ArunJ98:
This PR allows students to request a plan to be approved. The advisor then see's this request in their homepage for them to be able to review the plan before an upcoming appointment.


api/app/controllers/appointments_controller.rb
```diff
@@ -0,0 +1,46 @@
+class AppointmentsController < ApplicationController
+    before_action :set_advisor, :set_user
+
+    def index
+        unless authorized
+            render json: { error: "Requester is not an advisor" }, status: :bad_request
+            return
+        end
+
+        @appointments = Appointment.where(user_id: @advisor.id, seen: FALSE).where('appointment_time >= ? ', Date.today).order('appointment_time ASC').limit(10)
```
andrewjleung, line 10: Why have the limit here? Will this be paginated?

api/app/views/appointments/show.json.jbuilder
```diff
@@ -0,0 +1,3 @@
+json.appointment do |json|
+  json.partial! 'appointments/appointment', appointment: @appointment
+end
```
andrewjleung, line 3: newline

frontend/src/services/AppointmentService.ts
```diff
@@ -1,76 +1,83 @@
...
-      "major": "Computer Science",
-      "planId": 1,
-      "planName": "A really really really long name",
-      "planMajor": "Computer Science",
-      "appointmentTime": "Mar 14, 2021"
-    },
-  ];
-};
+  }).then(response => response.json());
+  // return [
```
andrewjleung, line 10: Let's move this commented mock data to `frontend/src/data/mockData.ts` instead of keeping it commented here 👍

andrewjleung (changes requested):
Few really small things but this looks great Arun!

api/app/controllers/appointments_controller.rb
ArunJ98: Definitely want to add pagination down the line, but didn't want to send too many appointments over in one request. Figured that if they can see the next X, those are the most important. Open to ideas here!

andrewjleung: If we're going to eventually add pagination then I think this is fine 👍

andrewjleung (approved):

PR #338, opened 2021-04-11
Proof-of-concept to add useForm hook for handling form state and validation AddPlanPopper

andrewjleung:
## Why

https://trello.com/c/i3oYMCc6

Making forms is hard. Making forms with validation is harder. Abstracting logic for validating forms is... ☹️ 

All of our forms that cleanly validate errors are difficult to set up and maintain, and there is a lot of common logic repeated across them. It would be nice to have some of this abstracted away to make things a little less painless or at least slightly more organized and readable. 

I ultimately decided not to pull in a library that handles this because it would probably just come with way more than we actually need and require too much setup than is worth, plus we'd need to pull in another whole bunch of dependencies when we could just make our own in-house lightweight alternative.

## What 

Questionably improves our form validation by adding a `useForm` hook. It abstracts out some simple core form validation functionalities that were used across the existing forms in the app such as validating fields, showing errors on submit, and resetting the form, while also containing the state for both fields and errors. This has just been added to the `AddPlanPopper` so far just to demonstrate it working.

Please destroy this in review so I can make it better/easier to use.

### DIFF
--- a/frontend/src/home/AddPlanPopper.tsx
+++ b/frontend/src/home/AddPlanPopper.tsx
@@ -44,7 +44,7 @@ import { ExcelUpload } from "../components/ExcelUpload";
 import { NextButton } from "../components/common/NextButton";
 import { RedColorButton } from "../components/common/ColoredButtons";
 import { SaveInParentConcentrationDropdown } from "../components/ConcentrationDropdown";
-import { useForm } from "../hooks/useForm";
+import { FormErrors, useForm } from "../hooks/useForm";
 
 const EXCEL_TOOLTIP =
   "Auto-populate your schedule with your excel plan of study. Reach out to your advisor if you don't have it!";
@@ -128,21 +128,11 @@ const INIT_FIELDS: AddPlanPopperFields = {
   basePlan: null,
 };
 
-const INIT_ERRORS = {
-  planName: "",
-  catalogYear: "",
-  major: "",
-  concentration: "",
-  coopCycle: "",
-  planOption: "",
-  basePlan: "",
-};
-
 const validateValues = (
   scheduleNames: string[],
   selectedDNDSchedule: React.MutableRefObject<DNDSchedule | undefined>,
   noConcentrationError: boolean
-) => (values: AddPlanPopperFields): typeof INIT_ERRORS => {
+) => (values: AddPlanPopperFields): FormErrors => {
   const validatePlanName = (values: AddPlanPopperFields) => {
     // Missing plan name
     if (!values.planName) {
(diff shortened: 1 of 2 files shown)


frontend/src/home/AddPlanPopper.tsx
```diff
@@ -95,19 +95,6 @@ const FieldContainer = styled.div`
...
-  margin: 8px;
-  flex-direction: row;
-  justify-content: center;
-`;
-
-const ErrorText = styled.p`
-  color: red;
-  text-align: center;
-`;
-
```
andrewjleung, line 110: These were unused

frontend/src/home/AddPlanPopper.tsx
```diff
@@ -121,52 +108,121 @@ interface Props {
...
-  const [selectedConcentration, setSelectedConcentration] = useState<
-    string | null
-  >(null);
-  const [selectedCoopCycle, setSelectedCoopCycle] = useState<string | null>(
-    null
-  );
-  const [selectedPlanOption, setSelectedPlanOption] = useState<string | null>(
-    null
-  );
-  const [selectedUserPlan, setSelectedUserPlan] = useState<string | null>(null);
```
andrewjleung, line 154: All this state now is within a single state object that is contained within the `useForm` hook.

frontend/src/home/AddPlanPopper.tsx
```diff
@@ -176,154 +232,96 @@ function AddPlanPopperComponent(props: Props) {
...
-      setSelectedCatalogYear(plan.catalogYear);
-      setSelectedMajor(
-        findMajorFromName(plan.major, allMajors, plan.catalogYear) || null
-      );
-      setSelectedConcentration(plan.concentration || null);
-      setSelectedCoopCycle(plan.coopCycle || "");
-      selectedDNDSchedule.current = plan.schedule;
-      counter.current = plan.courseCounter;
-    }
-  }, [selectedUserPlan, userPlans, allMajors]);
```
andrewjleung, line 193: This just got moved closer to where it was relevant

frontend/src/hooks/useForm.ts
```diff
@@ -0,0 +1,55 @@
...
+
+type Validator<V> = (values: V) => FormErrors;
+
+/**
+ * Custom hook for setting up a form with validation.
+ *
+ * @param initValues the initial values for each field
+ * @param validators the validator for the form
+ */
+export const useForm = <V>(initValues: V, validator: Validator<V>) => {
```
andrewjleung, line 15: Setting up a form requires specifying the initial values for each field in the form, along with a single validator function for creating an error object that contains the errors for each field.

The hook will give back several values which are used to then wire up the form, including the field values, a function for setting the values, the errors, a function for checking if there are any errors at all (used on submit), whether the errors should be visible, and a function for resetting the entire form.

imanmoreira (approved):
This looks amazing!

PR #339, opened 2021-04-18
Ensure that all co-op cycles are used/validated when creating from existing plans

andrewjleung:
https://trello.com/c/X7SrCXb9

## Why

There was another place where example schedules were made based upon co-op cycle and major on the `AddPlanPopper` which wasn't accounted for. This area needs to both use the additional co-op cycles and make sure to only allow creating from example plans where applicable. 

## What

- Use an existing helper function for determining whether there's an existing example plan for a major/co-op cycle
- Use `BASE_FORMATTED_COOP_CYCLES` on the `AddPlanPopper`
- Add logic to the `AddPlanPopper` to only show the `EXISTING_PLAN` option if one exists for the selected major and co-op cycle

### DIFF
--- a/frontend/src/home/AddPlanPopper.tsx
+++ b/frontend/src/home/AddPlanPopper.tsx
@@ -17,13 +17,15 @@ import { connect } from "react-redux";
 import { AppState } from "../state/reducers/state";
 import { Dispatch } from "redux";
 import { Major, Schedule, ScheduleCourse } from "../../../common/types";
-import { findMajorFromName } from "../utils/plan-helpers";
+import {
+  findExamplePlanFromCoopCycle,
+  findMajorFromName,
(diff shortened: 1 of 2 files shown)


imanmoreira (commented):
Hey! Looks good overall, but it also needs to be fixed where the student's initial plan is created (on the onboardingInfo page)

imanmoreira (changes requested):
So this also doesn't seem to be totally fixing the issue. The main problem is when a user tries to create a new plan with a coop cycle that does not have a corresponding example schedule, it is unable to create one. Here's an example of the issue:
image

justineluo:
[Bug fixes for Coop cycle change](https://trello.com/c/X7SrCXb9/325-bug-fixes-for-coop-cycle-change)

PR #361, opened 2021-10-30
Rename master to main in redeploy.sh

aryanshah701:
We renamed the master branch to main so need to rename it in the redeploy script too :)


justineluo (approved):
LGTM!

PR #362, opened 2021-11-02
Fix CI Pipeline

aryanshah701:
Node compatibility tests were failing because scraper tests were failing. The scraper tests were failing because some of the course catalog urls we were testing were now invalid so I updated them to the correct ones.


justineluo:
Fixed in another PR

PR #363, opened 2021-11-10
Add new major types

jia614:
Implemented new Major2 type :)

### DIFF
--- a/common/types.ts
+++ b/common/types.ts
@@ -153,6 +153,109 @@ export interface INEUPrereqCourse {
   missing?: true;
 }
 
+//                                       NEW MAJOR OBJECT HERE
+/**
+ * A Major, containing all the requirements.
+ * @param name The name of the major.
+ * @param requirementSections A list of the sections of requirements.
+ * @param totalCreditsRequired Total credits required to graduate with this major.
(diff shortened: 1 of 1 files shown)


common/types.ts
```diff
@@ -153,6 +153,109 @@ export interface INEUPrereqCourse {
...
+
+/**
+ * A Section, containing its related requirements.
+ * @param title The title of the section.
+ * @param requirements A list of the requirements within this section.
+ * @param minRequirementCount The minimum number of requirements (counts from requirements)
+ *        that are accepted for the section to be fulfilled.
+ */
+export interface Section {
+  title?: string;
```
cindy1u0, line 181: lets make this title: string | null

cindy1u0 (approved):
overall looks good to meeee :))))

common/types.ts
justineluo: Let's not make this optional, since I'm pretty sure every section needs a title. I updated the docs, sorry for the confusion!

justineluo (changes requested):
Just small fix, otherwise LGTM :)

justineluo (approved):

PR #371, opened 2021-12-09
await nested promise

justineluo:
We were not able to fetch data because of API change. In the screenshot we see the "data" field is now a promise instead of the major objects. This PR adds a "then" so that it waits for data to resolve. 

Screen Shot 2021-12-09 at 6 50 39 PM


cindy1u0 (approved):
LGTM xD thank you for fixing it :DDD

PR #375, opened 2022-01-02
Frontend react error handling

jia614:
frontend-errors

# Description
Refactor `ErrorPage` to `GenericErrorPage`, `FrontendErrorPage`, and `BackendErrorPage` for components for error catching.

## Type of change
Feature

# How Has This Been Tested?
Broke the frontend code and saw the error page pop up :') see gif ^
*Note: actual error pops up in dev mode on chrome, can be closed by hitting X or ESC, would not show up in prod

# Checklist:

### DIFF
--- a/frontend/src/error/ErrorHandler.tsx
+++ b/frontend/src/error/ErrorHandler.tsx
@@ -1,13 +1,15 @@
 import React from "react";
 import { useLocation } from "react-router-dom";
-import { get } from "lodash";
 import { BackendErrorPage } from "./ErrorPages";
 
-const ErrorHandler = ({ children }: any) => {
-  const location = useLocation();
-  const status: number = get(location.state, "errorStatusCode");
-  if (status >= 400) {
-    return <BackendErrorPage statusCode={status} />;
+type LocationState = {
+  errorStatusCode: number;
+};
+
+const ErrorHandler: React.FC = ({ children }: any): JSX.Element => {
+  const errorStatusCode = useLocation<LocationState>().state?.errorStatusCode;
+  if (errorStatusCode >= 400) {
+    return <BackendErrorPage statusCode={errorStatusCode} />;
   } else {
     return children;
   }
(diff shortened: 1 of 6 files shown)


frontend/src/error/ErrorHandler.tsx
```diff
@@ -1,13 +1,13 @@
 import React from "react";
 import { useLocation } from "react-router-dom";
 import { get } from "lodash";
-import { ErrorPage } from "./ErrorPage";
+import { BackendErrorPage } from "./ErrorPages";
 
 const ErrorHandler = ({ children }: any) => {
```
aryanshah701, line 6: We should probably type the ErrorHandler as `React.FC` so that we don't have to explicitly type the props or the return type of the component

frontend/src/error/ErrorHandler.tsx
```diff
@@ -1,13 +1,13 @@
...
   const location = useLocation();
```
aryanshah701, line 7: Hmm... curious if we can harness typescript to avoid spelling issues with "errorStatusCode" by typing the location state.

Does the following work by any chance?
```tsx
type LocationState = {
    errorStatusCode: number;
}

const { state } = useLocation<LocationState>()
const { errorStatusCode } = state;
```

frontend/src/error/GenericErrorPage.tsx
```diff
@@ -75,7 +75,7 @@ const ButtonWrapper = styled.div`
   width: 100%;
 `;
 
-export const ErrorPage = ({ statusCode }: { statusCode: number }) => {
+export const GenericErrorPage = ({ message }: { message: string }) => {
```
aryanshah701, line 78: Ooh I find typing function components to be cleaner with the `React.FC\<PropsType\>`, since it prevents both having inline types where we are destructuring props and having to specify a return type for the component. 

Nothing too wrong with this though! I would however like to see a return type for the component(`JSX.Element` I think). I find the return type for a component to be useful in cases where I do dumb stuff and either forget to return JSX or return something else lmao.

aryanshah701 (approved):
Looks solid!! Just a few smol comments :)

PR #378, opened 2022-01-25
Feat nest backend

aryanshah701:
# Description

A version 2 of the GraduateNU API in TypeScript and Nestjs. 

Closes [Try out nestjs backend](https://trello.com/c/4mJbmpC1/361-try-out-nextjs-backend)

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# How Has This Been Tested?

- Postman testing
  - Run `yarn install` to fetch dependencies
  - Ensure a community edition of PostgreSQL is running
  - Run `yarn run start:dev` to start up the server
  - Test the server through postman

# Checklist:


bugsalexander:
[Try out NestJS backend](https://trello.com/c/4mJbmpC1/361-try-out-nestjs-backend)

api-v2/src/student/dto/create-student.dto.ts
```diff
@@ -0,0 +1,70 @@
...
+  @IsObject()
+  coursesCompleted?: ScheduleCourse[];
+
+  @IsOptional()
+  @IsObject()
+  coursesTransfered?: ScheduleCourse[];
+
+  @IsOptional()
+  @IsInt()
+  primaryPlanId?: number;
```
cindy1u0, line 65: Do we assign a primary ID when a student is created?

scrapers/test/warnings.test.js
```diff
@@ -280,15 +268,12 @@ test("Range section warning produced for cs_pos_2.json", () => {
 
 test("Requirement group warnings produced appropriately for cs_sched_1.json", () => {
   let cs_sched = fs.readFileSync(
-    "./scrapers/test/mock_schedules/cs_sched_1.json",
+    "test/mock_schedules/cs_sched_1.json",
```
cindy1u0, line 271: yesss make CI/CD WORKKK

api-v2/src/student/dto/create-student.dto.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  @IsNotEmpty()
+  @IsString()
+  password: string;
+
+  @IsOptional()
+  @IsInt()
+  @Min(1898) // the year NEU was established
+  @Max(3000) // will GraduateNU last beyond year 3000?!
+  academicYear?: number;
```
cindy1u0, line 33: I am little concerned or I am just thinking if we should make academic graduateYear and stuff optional here. I know that for our current implementation a student signs up first, and when that happens these fields are all optional. I am just a bit concerned that if frontend doesn't do a good job enforcing onboarding, then these values will be null, and it will potentially break the home plan page. Any thoughts on this?

cindy1u0 (approved):
U R CRACKED HOLY SHIT. IT'S SO CLEAN IM CRYING ;-;-;-;-;-;-;-;-;-;  THANKS FOR YOUR HARD WORK!!! Also i have two questions
1. `IsWithPlan` -  I see that it's finding the plan relation. Does it determine whether or not to show the plans of the student or whether or not to show students without plans?
2. (Nit) Some of the error messages has an additional comma after the ID lol. It looks something like "A Student with the given id, is not found."

cindy1u0:
tested with postman test suite - things seemed to work fine!! Tried some corner cases and all seemed to work too!!

scrapers/test/warnings.test.js
aryanshah701: ahahahah yasss!!!

api-v2/src/plan/entities/plan.entity.ts
```diff
@@ -0,0 +1,57 @@
...
+      onDelete: 'CASCADE',
+    },
+  )
+  student: Student;
+
+  @Column({ type: 'json' })
+  schedule: Schedule;
+
+  @Column()
+  major: string;
```
bugsalexander, line 35: it's possible for a student to have multiple majors/minors, right?

api-v2/src/plan/entities/plan.entity.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  @Column()
+  coopCycle: string;
+
+  @Column({ nullable: true })
+  concentration: string;
```
bugsalexander, line 41: same w above – perhaps this should be attached to the "major"

api-v2/src/plan/entities/plan.entity.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  @Column({ type: 'smallint', default: 2018 })
+  catalogYear: number;
```
bugsalexander, line 44: why default?

api-v2/src/plan/entities/plan.entity.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  @Column({ type: 'json' })
+  courseWarnings: CourseWarning[];
+
+  @Column({ type: 'json' })
+  warnings: IWarning[];
```
bugsalexander, line 50: i think it's fine to omit these, since we only calculate them on the frontend – does the current backend save them?

api-v2/src/plan/plan-not-found.error.ts
```diff
@@ -0,0 +1,5 @@
+export class PlanNotFoundError extends Error {
+  constructor(planId: number) {
+    super(`A Plan with the id, ${planId}, is not found`);
+  }
```
bugsalexander, line 4: should we instead return a ForbiddenError? theoretically the OwnPlanGuard should protect against users accessing plans that aren't theirs, but just in case

api-v2/src/plan/plan.controller.ts
```diff
@@ -0,0 +1,74 @@
...
+  @Post()
+  async create(
+    @Body() createPlanDto: CreatePlanDto,
+    @Req() req: AuthenticatedRequest,
+  ): Promise<Plan> {
+    try {
+      return await this.planService.create(createPlanDto, req.user);
+    } catch (error) {
+      throw new BadRequestException(error.message);
+    }
```
bugsalexander, line 39: save exceptions for exceptional behavior, in which case prefer 500. if it's valid for this to throw, then perhaps make its return type different

api-v2/src/plan/plan.module.ts
```diff
@@ -0,0 +1,12 @@
+import { Module } from '@nestjs/common';
+import { PlanService } from './plan.service';
+import { PlanController } from './plan.controller';
+import { Plan } from './entities/plan.entity';
+import { TypeOrmModule } from '@nestjs/typeorm';
+
+@Module({
+  imports: [TypeOrmModule.forFeature([Plan])],
```
bugsalexander, line 8: this is cool ,what does it do? (curious)

api-v2/src/plan/plan.service.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  create(createPlanDto: CreatePlanDto, student: Student): Promise<Plan> {
+    const newPlan = this.planRepository.create({ ...createPlanDto, student });
+
+    try {
+      return this.planRepository.save(newPlan);
+    } catch (_error) {
+      throw new Error('Something went wrong when creating the new Plan');
+    }
+  }
```
bugsalexander, line 26: when would the save ever throw?

api-v2/src/plan/plan.service.ts
```diff
@@ -0,0 +1,77 @@
...
+    planId: number,
+    loggedInStudent: Student,
+  ): Promise<boolean> {
+    try {
+      const { student: planOwner } = await this.findOne(planId);
+      return StudentService.compareStudents(planOwner, loggedInStudent);
+    } catch (_error) {
+      /**
+       * findOne failed probably since the plan doesn't exist. A plan that doesn't
+       * exist isn't owner by the anyone so returning false.
```
bugsalexander, line 51: does it not return null?

api-v2/src/plan/plan.service.ts
```diff
@@ -0,0 +1,77 @@
...
+    }
+  }
+
+  async update(id: number, updatePlanDto: UpdatePlanDto) {
+    const updateResult = await this.planRepository.update(id, updatePlanDto);
+
+    if (updateResult.affected === 0) {
+      // no plan was updated, which implies a plan with the given id wasn't found
+      throw new PlanNotFoundError(id);
+    }
```
bugsalexander, line 63: seeing all these error cases is making me want to write test cases to cover them all lol

api-v2/src/student/dto/create-student.dto.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  @IsOptional()
+  @IsInt()
+  @Min(1898)
+  @Max(3000)
+  catalogYear?: number;
+
+  @IsOptional()
+  @IsString()
+  major?: string;
```
bugsalexander, line 49: multiple majors?

api-v2/src/student/entities/student.entity.ts
```diff
@@ -0,0 +1,79 @@
...
+  CreateDateColumn,
+  UpdateDateColumn,
+  OneToMany,
+} from 'typeorm';
+import * as bcrypt from 'bcrypt';
+import { Exclude } from 'class-transformer';
+import { ScheduleCourse } from '../../../../common/types';
+import { Plan } from '../../plan/entities/plan.entity';
+
+@Entity()
```
bugsalexander, line 15: it's confusing what types are db types and what are not. what KOH did was put the DTO types in `common` package (shared between frontend and backend) and then put db entities in .entity.ts files (in the relevant module folder in backend), see https://github.com/sandboxnu/office-hours/blob/master/packages/server/src/course/course.entity.ts

api-v2/src/student/student.controller.ts
```diff
@@ -0,0 +1,123 @@
...
+import { JwtAuthGuard } from 'src/guards/jwt-auth.guard';
+import { DevRouteGuard } from 'src/guards/dev-route.guard';
+import { AuthenticatedRequest } from 'src/auth/interfaces/authenticated-request';
+
+@Controller('students')
+export class StudentController {
+  constructor(private readonly studentService: StudentService) {}
+
+  @UseGuards(JwtAuthGuard)
+  @Get('/me')
```
bugsalexander, line 31: is this route supposed to be `.../students/me`? if so then the prefix / is unnecessary

api-v2/src/student/student.controller.ts
```diff
@@ -0,0 +1,123 @@
...
+  async getMe(
+    @Req() req: AuthenticatedRequest,
+    @Query('isWithPlans', new DefaultValuePipe(false), ParseBoolPipe)
```
bugsalexander, line 34: what does this do?

api-v2/src/student/student.module.ts
```diff
@@ -0,0 +1,13 @@
+import { Module } from '@nestjs/common';
+import { StudentService } from './student.service';
+import { StudentController } from './student.controller';
+import { TypeOrmModule } from '@nestjs/typeorm';
+import { Student } from './entities/student.entity';
+
+@Module({
+  imports: [TypeOrmModule.forFeature([Student])],
```
bugsalexander, line 8: rather than having to put this everywhere, could alternatively do it once in app.module.ts if you don't mind not specifying specific feature, [see here](https://github.com/sandboxnu/office-hours/blob/cca52eb54092abe674fb53f3086453bc61613b94/packages/server/src/app.module.ts#L26)

api-v2/src/student/student.service.ts
```diff
@@ -0,0 +1,108 @@
...
+} from 'typeorm';
+import { CreateStudentDto } from './dto/create-student.dto';
+import { UpdateStudentDto } from './dto/update-student.dto';
+import { Student } from './entities/student.entity';
+import { StudentNotFoundError } from './student-not-found.error';
+
+@Injectable()
+export class StudentService {
+  constructor(
+    @InjectRepository(Student)
```
bugsalexander, line 17: with above, this would then just be [this](https://github.com/sandboxnu/office-hours/blob/cca52eb54092abe674fb53f3086453bc61613b94/packages/server/src/course/course.service.ts#L29) – fewer params/code, but less specific. up to you

api-v2/src/student/student.service.ts
```diff
@@ -0,0 +1,108 @@
...
+    private studentRepository: Repository<Student>,
+  ) {}
+
+  async create(createStudentDto: CreateStudentDto): Promise<Student> {
+    // make sure the user doesn't already exists
+    const { email } = createStudentDto;
+    const userInDb = await this.studentRepository.findOne({ where: { email } });
+    if (userInDb) {
+      throw new Error('A user with the email is already registered');
```
bugsalexander, line 26: we should put error messages into a constant, just so changing them is easier later on. something like ERROR_MESSAGES.controllers.student.alreadyExists. 

also here, why throw Error, instead of HttpError? i saw in earlier places in code, try/catch around calling service methods and then throwing HttpError (assumes service just throws normal Errors), but also some throwing HttpError in (other) service methods (assumes no try/catch around service method call). we should pick one convention, and then stick with it. two good options:

- ignore "controllers are http only layer" and just throw HttpError in services, always–this would then mean no try/catches around service method calls, since we have to propagate the HttpErrors to the controller
- be good citizens and customize return values; save exceptions for exceptional behavior (don't do throwing in services generally, just return null/something to signify what happened), then throw HttpError/handle it in controller. probably more work, but probably cleaner as well.

api-v2/src/student/student.service.ts
```diff
@@ -0,0 +1,108 @@
...
+
+    const newStudent = this.studentRepository.create(createStudentDto);
+    try {
+      return this.studentRepository.save(newStudent);
+    } catch (error) {
+      throw new Error(error.message);
+    }
+  }
+
+  findAll(): Promise<Student[]> {
```
bugsalexander, line 37: im too lazy to lookup where this is used, but when would we ever need this? it should probably be paginated (there will potentially be thousands of students)

api-v2/src/student/student.service.ts
```diff
@@ -0,0 +1,108 @@
...
+    const deleteResult = await this.studentRepository.delete(uuid);
+
+    if (deleteResult.affected === 0) {
+      throw new StudentNotFoundError();
+    }
+
+    return deleteResult;
+  }
+
+  static compareStudents(student1: Student, student2: Student): boolean {
```
bugsalexander, line 105: nit: naming it `compare` implies that students are `Comparable` (i love my java) and that this should return a number. rename it to studentsEqual (or similar) if it's for checking equality (returning bool)

infrastructure/develop/docker-compose.yml
```diff
@@ -0,0 +1,16 @@
+version: "3"
+
+services:
+  postgres:
+    image: postgres:latest
```
bugsalexander, line 5: we should use same version as in prod... idk what prod will use, but koh uses 11.5 if we wanna use that

scrapers/src/json_loader.ts
```diff
@@ -28,7 +28,7 @@ const getClassMapLinkForTermId = (termId: number) => {
 // ./maps/${TERMID}.json
 const DIR = "maps";
 const getClassMapLinkLocal = (termId: number) => {
-  return `./${DIR}/${termId}.json`;
+  return `../${DIR}/${termId}.json`;
```
bugsalexander, line 31: holy shit is this gonna get used somewhere :eyes: (i wrote it like 3 years ago lol)

bugsalexander (commented):
whew. lots of comments lmao, so dont feel afraid to push back (i'm curious to see what u/others think). great work!

main points:

- we're preparing for yarn workspaces, don't worry about DTO clutter too much for now
- decide on convention for throwing HttpError in services or not
- standardize errors in some constant somewhere
- can make postgres connection global (avoid specifying in dependency in every service)
- plans/student data modelling might need more thoughts (multiple majors/concentrations?)
- can remove the node-compability check thing lol

.github/workflows/main.yml
```diff
@@ -17,6 +18,25 @@ jobs:
...
       - name: Run Jest Tests
         run: |
+          cd scrapers 
+          yarn install
+          yarn test
+  run-api-v2-tests:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node: ["12"]
```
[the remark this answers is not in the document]
bugsalexander: i also just realized this is only one value, so there's no need to specify a strategy matrix – just hardcode 12 in the node version step

api-v2/src/student/dto/create-student.dto.ts
bugsalexander: ignore, cindy told me we don't support

api-v2/src/student/dto/create-student.dto.ts
bugsalexander: i talked with cindy on slack a bit about this and i think in the future we should enforce onboarding on the frontend, and always assume user has graduation date on the backend (make it required)

for the case where onboarding gets interrupted (user reloads page/quits browser), we should save in localstorage between steps in the browser, and resume on frontend if necessary

bugsalexander:
overall SUPER chungus dude so incredible u were able to write this all over winter break!!!

api-v2/src/auth/auth.service.ts
```diff
@@ -0,0 +1,77 @@
...
+    // find student in db
+    const student = await this.studentService.findByEmail(email);
+
+    if (!student) {
+      throw new Error('Student with given email not found');
+    }
+
+    // validate creds
+    const { password: trueHashedPassword } = student;
+    const isValidPassword = await bcrypt.compare(password, trueHashedPassword);
```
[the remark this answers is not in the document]
aryanshah701: ah yep trueHashedPassword is hashed since it's coming from the db(we store hashed passwords in the db using a `@BeforeInsert` in the entity class). Does that answer the question?

api-v2/src/auth/dto/login-student.dto.ts
```diff
@@ -0,0 +1,11 @@
+import { IsEmail, IsNotEmpty, IsString } from 'class-validator';
+
+export class LoginStudentDto {
```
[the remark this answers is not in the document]
aryanshah701: for sure!! We'll def move everything into the common directory once we finish the file cleanup and yarn workspaces move!

api-v2/src/interceptors/transform-response-interceptor.ts
```diff
@@ -0,0 +1,32 @@
...
+export interface ServerResponse<T> {
+  statusCode: number;
+  data: T;
+}
+
+/**
+ * Transforms all outgoing data into the ServerResponse shape above.
+ */
+@Injectable()
+export class TransformResponseInterceptor<T>
```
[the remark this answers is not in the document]
aryanshah701: oop my bad the comment is outdated, i'll change it. but essentially it transforms all outgoing data into the `ServerDataResponse` type which essentially standardizes the shape of what our server responds with.

I did this cause i found it kinda odd to have our responses simply be the actual data. For instance the student get endpoint returns the student itself so a client would essentially do `response.catalogYear` etc. At the same time I didn't want to manually go and return a specific response shape from each controller method. So this global interceptor takes the data returned from controllers and transforms it into a standardized recognizable shape -
```
{
  data: T,
  statusCode: number
}
```
This meshes kinda well with the `ServerErrorResponse` type(which is what nest returns if we throw an HttpError). So our frontend can safely expect either a ServerErrorResponse or a ServerDataResponse. We can then use a type guard to handle both cases on the frontend with full type safety. Essentially the frontend would be forced to handle the error case before proceeding to handle the data case.

All the above types are at the top of the common/types file!

Does this make sense? Would love to know your thoughts on this, it's just a design decision I made so is definitely arguable! Would also

api-v2/src/plan/entities/plan.entity.ts
aryanshah701: Yep the current backend does save them. I kinda just copy pasta'd the entity shapes but there's no reason as to why we shouldn't revise them. I don't see any particular reason as to why we should keep them so I can strip them out. We can always bring them back if we need to? Any thoughts?

api-v2/src/plan/plan-not-found.error.ts
aryanshah701: So I throw this error specifically when the plan id literally doesn't exist in the db. This is a generic error thrown by services, https agnostic.

As for whether we should throw Forbidden in controllers when a plan doesn't exist, I think there is merit in doing so.

The sort of specificity we get from throwing a NotFound when a plan doesn't exist and Forbidden when a user is not authorized may make things less secure(since now the client knows if the id doesn't exist or if they are forbidden from accessing it).

At the same time the specificity allows us to handle the case where somehow a user that used to own a plan but deleted it accesses that plan. In a case like that we might want to display "Oh no, that plan doesn't exist anymore" rather than "Forbidden"? 

I could make an argument for handling it either way at the controller level, thoughts?

api-v2/src/plan/plan.module.ts
aryanshah701: the repository pattern - informs nest to just make the Plan repository available in providers of the module!

api-v2/src/plan/plan.service.ts
aryanshah701: save would throw if `newPlan` violates any table constraint(duplicate key, unique field, non nullable) etc.

api-v2/src/plan/plan.service.ts
aryanshah701: lmaoo yea, writing endpoint tests shouldn't be too much work. I've manually tested every edge case and that definitely isn't the way to do things lol.

Maybe we can make this a part of this PR or split it out into its own ticket if someone wants to gain some experience with nest, jest, and supertest. The good thing is that we're merging into a feature branch and not main. But if you'd like to see them in this PR, can totally make it happen!

api-v2/src/student/entities/student.entity.ts
aryanshah701: So I agree with moving DTO types into the common package after yarn workspace move! 

But I didn't quite understand why it was difficult to see which types were entity and which weren't since entity files are named `x.entity.ts` in the `entity` directory and DTOs are `x.dto.ts` in the `dto` directory. 

Did you mean that we should move the `.entity.ts` files out of the `entity` directory and into the module directory?

api-v2/src/student/student.controller.ts
aryanshah701: ahh yes, thanks for catching this!

api-v2/src/student/student.service.ts
aryanshah701: not used anywhere! just used for dev routes that I used for debugging. I'll just remove the dev routes + any service methods associated to them to avoid confusion

infrastructure/develop/docker-compose.yml
aryanshah701: totally!!

scrapers/src/json_loader.ts
aryanshah701: ahahah na, we re-organized the tests a bit in CI and needed to make this update to make them pass lmao

api-v2/src/interceptors/transform-response-interceptor.ts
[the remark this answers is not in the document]
bugsalexander: Makes sense, but I think including the http code in the body is redundant.

on the frontend, we can explicitly handle all non GET request responses anyway, so I think it's fine to not include. Thoughts?

api-v2/src/plan/entities/plan.entity.ts
bugsalexander: To add on to this, i think we should have no default here-it should be an error for the catalog year to not be specified, and i dont think 2018 makes sense as a universal default

api-v2/src/plan/entities/plan.entity.ts
bugsalexander: Sounds good to remove 👍

api-v2/src/plan/plan-not-found.error.ts
bugsalexander: Ohh i see, thanks. Yea ignore this

api-v2/src/student/entities/student.entity.ts
bugsalexander: Ohhb LOL ignore me haha i didnt see file extension

api-v2/src/student/student.service.ts
bugsalexander: Nah i think it's fine to keep dev routes, but deleting unised service methods sounds good to me

api-v2/src/student/dto/create-student.dto.ts
aryanshah701: hmm we don't assign a primaryPlanId initially since the student doesn't have a plan when created. this again has to do with how our onboarding works so we can change when we figure out onboarding!

api-v2/.env.development
```diff
@@ -0,0 +1,4 @@
+DB_USERNAME=postgres
+DB_URL=postgres://postgres@localhost:5432/graduatenu_api_dev
+JWT_SECRET_KEY=chungusdevs
+JWT_EXPIRE_TIME=86400000
```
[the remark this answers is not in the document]
aryanshah701: a 100% lmao, i forgot we could add comments to .envs

api-v2/README.md
```diff
@@ -0,0 +1,73 @@
...
+```bash
+# development
+$ npm run start
+
+# watch mode
+$ npm run start:dev
+
+# production mode
+$ npm run start:prod
+```
```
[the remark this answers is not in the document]
aryanshah701: oop yeap i'll make them `yarn start`. that's what you meant tho right?

.github/workflows/main.yml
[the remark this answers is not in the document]
aryanshah701: the latest stable version of node is actually 16.14, should we use that instead? I also just realize i was using v16 in development lol

api-v2/src/auth/auth.controller.ts
```diff
@@ -0,0 +1,36 @@
...
+import { CreateStudentDto } from 'src/student/dto/create-student.dto';
+import { Student } from 'src/student/entities/student.entity';
+import { AuthService } from './auth.service';
+import { LoginStudentDto } from './dto/login-student.dto';
+
+@Controller('auth')
+export class AuthController {
+  constructor(private readonly authService: AuthService) {}
+
+  @Post('register')
```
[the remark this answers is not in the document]
aryanshah701: oop sorry I didn't get you, a pipe to validate what? The body is validated through the global pipe and class-validator 👀

api-v2/src/interceptors/transform-response-interceptor.ts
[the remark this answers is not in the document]
aryanshah701: yeap no reason to handle the statusCode so I can strip that off!

api-v2/README.md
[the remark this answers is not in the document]
bugsalexander: yep!

api-v2/src/plan/entities/plan.entity.ts
aryanshah701: 100%. lowkey i just copy pasta'd the entity types from the previous backend and the previous backend had this for some odd reason. Didn't really use many brain cells when defining the entities haha.

.github/workflows/main.yml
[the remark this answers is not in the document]
bugsalexander: LOL good question – up to you!

api-v2/src/plan/plan.controller.ts
aryanshah701: Makes sense! also sorry, what do you mean by make its return type different? Do you mean make the return type of the service different(i.e undef if the creation failed and Plan if it didn't)?

Thought process behind this was that the `create` probably failed since the client request body had issues with table constraints we have. But you're right in that the create could also straight up fail for other 500 reasons(typeorm/db issues).

api-v2/src/plan/plan.service.ts
aryanshah701: oh yea my bad!!

api-v2/src/student/student.controller.ts
aryanshah701: Haha not sure if this is about the "isWithPlan" search param or the default value pipe or the parse bool pipe so imma just explain all 3.
- isWithPlan is a boolean search param that tells us whether we need to return a student with their plans or not. Implemented this so that we can don't always perform table joins when providing students unless we need. Not sure if this is useful yet(maybe for pages where we just display username or something?) but I just default to providing an option like this when dealing with multiple tables. This could be argued so feel free to if you have thoughts!
- DefaultValuePipe(false) makes the param `false` by default if it isn't specified
- ParseBoolPipe extracts the string for the search param and makes sure it is "true" or "false". If responds with an error saying `isWithPlans` should be "true" or "false". If it is a valid string then it parses it into a boolean depending on the string.

aryanshah701: Did any of the above 3 things answer the question haha?

api-v2/src/student/student.module.ts
aryanshah701: From the file you linked i think we have it setup similarly as well in app.module(correct me i'm wrong though!). The `connection` object is available throughout the application without requiring feature imports but I find the repository pattern much better because 
- it explicitly requires us to think about what tables each module interacts with 
- kinda follows the least privilege access principle. using the entire typeorm connection object is way too much functionality that most providers don't need. we mostly simply need the repository(s) for the table(s) we are accessing/updating
- i tend to like the repository pattern in general as a good way of using typeorm in any project since it abstracts away unnecessary db terms like `connection` and gives us a nice/simple interface that lets us do things with a specific table
Do the points make sense?

api-v2/src/student/student.service.ts
aryanshah701: I like both these options, would like to discuss them a little more during GM!

My current implementation has 2 intentions
- Catching issues in a method as soon as possible and throwing an error instead of relying on classes using the method to throw errors
- Not throwing HttpError in service

I may have by mistake thrown an HttpError in a service but i can't seem to find it but that definitely isn't my intention.

api-v2/src/student/student.service.ts
aryanshah701: ah cool. well we'll have to keep this one since it's used by a dev route... i'll do a pass and make sure we don't have unused service methods too!

api-v2/src/interceptors/transform-response-interceptor.ts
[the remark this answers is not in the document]
bugsalexander: and then i guess after that, there's no point in wrapping `data` in an object, because axios does this already, so i think it would be good to remove the entire wrapper

api-v2/src/plan/entities/plan.entity.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  @Column()
+  name: string;
+
+  @ManyToOne(
+    () => Student,
+    student => student.plans,
+    {
+      cascade: true,
+      onDelete: 'CASCADE',
```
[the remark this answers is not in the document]
bugsalexander: yeah that's true. your call! I think it would be acceptable to either cascade, or just have two calls to `delete` in the handler (one for student, one for `delete from plans where student_id = <id>`)

api-v2/src/plan/entities/plan.entity.ts
bugsalexander: LOL npnp

api-v2/src/plan/plan.controller.ts
bugsalexander: yeah exactly. in general with JS/TS i prefer to not use try/catch because there aren't ways to match on what was thrown. for example, did we throw a normal `Error`, or a `PlanNotFound` error?

and yea sorry by "make return type different" i meant encode the possible failure modes into the type signature, so we can be sure that they are handled. this is nice, because then in the controller layer we can just throw the corresponding HttpError (or subclass). and then if something unexpected throws, it will get propagated automatically since we aren't try/catching everything. 

the only thing is there isn't a clean way to do this in typescript :((((

api-v2/src/plan/plan.service.ts
bugsalexander: nah we do this later 🤪

api-v2/src/student/student.controller.ts
bugsalexander: yep!

api-v2/src/plan/plan.controller.ts
bugsalexander: discussed offline: decided to never throw in services (do try/catch)

api-v2/src/plan/entities/plan.entity.ts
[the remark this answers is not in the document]
bugsalexander: either ok

api-v2/src/auth/jwt.strategy.ts
```diff
@@ -29,10 +29,12 @@ export class JwtStrategy extends PassportStrategy(Strategy) {
    * @param jwtPayload decoded payload stored in the JWT
    */
   async validate(jwtPayload: JwtPayload): Promise<Student> {
-    try {
-      return await this.authService.validateJwtPayload(jwtPayload);
-    } catch (_error) {
+    const student = await this.authService.validateJwtPayload(jwtPayload);
+
+    if (!student) {
       throw new UnauthorizedException('Invalid JWT');
```
bugsalexander, line 35: thoughts on returning without a string here? i'm cautious of having good error messages for login, because of information it might leak to hackers

api-v2/src/plan/plan.service.ts
```diff
@@ -21,7 +20,7 @@ export class PlanService {
     try {
       return this.planRepository.save(newPlan);
     } catch (_error) {
-      throw new Error('Something went wrong when creating the new Plan');
+      return null;
```
bugsalexander, line 23: like you mentioned when we talked in person, here we lose information, and so it would be good to log this (sentry or something) later on. can you add a todo?

bugsalexander (commented):
diff looks good so far! i'll do a once over then good to go

api-v2/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+- If you want to change the way things work, start a discussion, make the change and document it here!
+
+- If you're making changes to the codebase, please ensure that your changes follow the design documented here!
+
+### Error Handling
+
+- Services will be HTTP agnostic and not throw HTTP errors.
+
+- Services will not throw errors in general, and will return `null` for exceptional behavior. For instance, if a user with the same email already exists, return `null` and expect the controller to handle this.
```
bugsalexander, line 40: instead of "_will return `null` for exceptional behavior_" i think it should say "_will not intentionally throw exceptions_" – in the future we might want to distinguish certain errors by a different return type, for example:

```typescript
// in types.ts
type Result<T, E> = 
  | { type: "ok", ok: T }
  | { type: "err", err: E }

type LoginError = {
  msg: string,
  type: "baduser" | "badpass" | ...
}

// in code
let result: Result<User, LoginError> = ...
if (result.type === "ok") {
  return result.ok;
}
switch (result.err.type) {
  case "baduser": return new Error(`bad user: ${result.err.msg}`);
  case "badpass": return new Error(`bad pass .... `);
  ...
}
```

api-v2/package.json
```diff
@@ -0,0 +1,90 @@
...
+    "start:prod": "NODE_ENV=production node dist/main",
+    "lint": "eslint \"{src,apps,libs,test}/**/*.ts\" --fix",
+    "test": "jest",
+    "test:watch": "jest --watch",
+    "test:cov": "jest --coverage",
+    "test:debug": "node --inspect-brk -r tsconfig-paths/register -r ts-node/register node_modules/.bin/jest --runInBand",
+    "test:e2e": "jest --config ./test/jest-e2e.json"
+  },
+  "dependencies": {
+    "@golevelup/ts-jest": "^0.3.2",
```
bugsalexander, line 24: goes in devDependencies i think

api-v2/src/interceptors/transform-response-interceptor.ts
```diff
@@ -0,0 +1,31 @@
+import {
+  CallHandler,
+  ExecutionContext,
+  Injectable,
+  NestInterceptor,
+} from '@nestjs/common';
+import { Observable } from 'rxjs';
+import { map } from 'rxjs/operators';
+
+export interface ServerResponse<T> {
```
bugsalexander, line 10: now that `ServerResponse` contains no additional information on top of the payload, could we remove it? it's annoying on the frontend to have to do `.data` a lot. axios by default adds a `.data`, so right now our frontend would look like this:

```typescript
const { data, error } = useSWR("/api/v2/students/1", (path) => axios.get(path));
const student = data.data.data;
```

api-v2/src/main.ts
```diff
@@ -0,0 +1,38 @@
...
+   * with the "@Exclude" decorator from being sent in the response.
+   * For this to work, returned objects in controllers have to be an actual instance
+   * of the type with the appropriate decorators.
+   */
+  app.useGlobalInterceptors(new ClassSerializerInterceptor(app.get(Reflector)));
+
+  /**
+   * Tranform all data returned from controllers to the ServerResponse shape.
+   */
+  app.useGlobalInterceptors(new TransformResponseInterceptor());
```
bugsalexander, line 33: could then delete this as well

bugsalexander (commented):
just a few more!

api-v2/src/auth/auth.service.ts
```diff
@@ -41,15 +45,15 @@ export class AuthService {
...
+      return null;
     }
 
     // validate creds
     const { password: trueHashedPassword } = student;
     const isValidPassword = await bcrypt.compare(password, trueHashedPassword);
 
     if (!isValidPassword) {
-      throw new Error('Invalid password');
+      return null;
```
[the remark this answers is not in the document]
aryanshah701: definitely!! even though it is a good thing to be ambiguous in this case, I do think that a more specific error handling solution for services will be nice in the future. I think services should be specific about the errors and controllers should make the decision of whether or not the HTTP error should be ambiguous 👀

api-v2/src/auth/jwt.strategy.ts
aryanshah701: Good point!

api-v2/README.md
aryanshah701: A 100%. I'll change the text but I definitely anticipate the design section for error handling to change when we implement more specific error handling like you suggested! For now I just wanted to document the design our codebase currently follows with the intention that when we update the design we will change the doc.

aryanshah701: I think I'm actually going to leave it as is since I think the doc should specifically represent the design the codebase currently follows. I'll leave a note for how we should probably make the error handling better though!

api-v2/src/interceptors/transform-response-interceptor.ts
aryanshah701: ooh yea i can remove it, fetch doesn't add the data section but we should probably be using `axios` anyways lmao

api-v2/package.json
aryanshah701: ah yess!!

api-v2/README.md
bugsalexander: I think "_will not intentionally throw exceptions_" still captures the same meaning, because returning null is the same as not intentionally throwing an exception?

bugsalexander (approved):
YEEET ty for answering all my review comments!

api-v2/README.md
aryanshah701: True true, but wanted to be more specific as to exactly what we do now and why it isn't great. That essentially explicitly calls out this somewhat jank way of handling errors, which explicitly provides an opportunity for improvement and hence a change to the error design section 🤔

PR #381, opened 2022-02-04
Create Semester Block Component

BrandonLim8890:
Semester Block Component

# Description

Created a semester block component to adhere to new designs.

## Type of change

# How Has This Been Tested?

Played around with UI to test functionality

# Checklist:


frontend/src/components/Year/Year.tsx
```diff
@@ -74,37 +87,55 @@ class YearComponent extends React.Component<Props> {
...
-            schedule={schedule}
-            isEditable={isEditable}
-            transferCourses={this.props.transferCourses}
-          />
-        </YearTopWrapper>
-        {isExpanded && (
-          <YearBody>
-            <SemesterBlock
+          <SemestersWrapper>
+            <SemesterContainer
```
BrandonLim8890, line 91: Want to find a way to map through the different semesters throughout the year instead of hardcoding each semester.

aryanshah701: Hmm, perhaps mapping through the semesters array? It seems like yall were on the path to doing something similar since I saw a semesters array in the `SemesterTop` component.

Something like this seemed to work for me

```tsx
const semesters: SeasonWord[] = ["fall", "spring", "summer1", "summer2"];
{semesters.map(semester => {
              const scheduleTerm = schedule.yearMap[year][semester];
              return (
                <SemesterContainer
                  isEditable={isEditable}
                  semester={scheduleTerm}
                  year={year}
                  semesterStatus={
                    scheduleTerm.status as StatusEnum
                  }
                  schedule={schedule}
                  transferCourses={this.props.transferCourses}
                  semesterWord={semester}
                  isExpanded={isExpanded}
                />
           );
})}
```

frontend/src/components/Year/Year.tsx
```diff
@@ -15,7 +15,12 @@ import { AppState } from "../../state/reducers/state";
 import { connect } from "react-redux";
 import { Dispatch } from "redux";
 import { toggleYearExpandedForActivePlanAction } from "../../state/actions/userPlansActions";
-import { ScheduleCourse } from "../../../../common/types";
+import {
+  ScheduleCourse,
+  SeasonWord,
+  StatusEnum,
+} from "../../../../common/types";
+import SemesterContainer from "../SemesterContainer";
```
aryanshah701, line 23: Totally not part of this ticket but it would be reallyy nice if we could turn this into a functional component 🥺 since we're trying to move towards those as and when we touch legacy class components

frontend/src/components/SemesterContainer.tsx
```diff
@@ -0,0 +1,44 @@
+import React from "react";
```
aryanshah701, line 1: Since we seem to have a ton of "Semester" related components, what do yall think about cleaning up this directory and moving them all into a "Semester" directory within components? 

I think the "Year" directory also has a  `SemesterType` component which could be moved into this new "Semester" directory.

aryanshah701 (changes requested):
Awesome job yall, and super cool to get your first PR out in like 5 days @BrandonLim8890!!!!
I've done a first pass and left a few comments, ping me once they're addressed and lets get this merged!

aryanshah701:
Oh another thing, lets get main merged in so that the node compatibility tests don't fail and there are no conflicts!

api/config/local_env (1).yml:Zone.Identifier
```diff
@@ -0,0 +1,4 @@
+[ZoneTransfer]
+ZoneId=3
+ReferrerUrl=https://drive.google.com/
+HostUrl=https://doc-0g-ac-docs.googleusercontent.com/docs/securesc/fhcaifmg0vulj6f9umn7j7ckq4c197uv/2phl10snt8j4viqpkta5pq7pdqkcuq26/1643936775000/01991989900526178905/13836624057235932916/1diBjgWU5k6l1zXtmahlEHOlnEfICh9QT?e=download&authuser=1
```
aryanshah701, line 4: ooh what does this file do?

aryanshah701 (commented):
Sweet, thanks for the quick changes yall!

Last 3 things before yall can merge
1. What are the 2 new files in the api directory?
2. I had a comment on this in my previous review - tappedSemester could be null according to the arg type definition but we're making the claim that is is definitely not null by putting the exclamation point. I had left 3 potential ways in my previous comment of dealing with it so lets get rid of that "!" :) 
3. I had also left a comment about `props.semester as SeasonWord` casting. It seems like yall marked that as resolved but the cast is still happening. Any reason we can't just change the prop type for `semester` to SeasonWord instead of string?

aryanshah701 (approved):
FIREEEEEE

PR #382, opened 2022-02-05
Hide feedback button

rael346:
# Description

This feature will hide or show the feedback button on the home page, depending on the environment variable REACT_APP_FEEDBACK_BUTTON. [Reference](https://create-react-app.dev/docs/adding-custom-environment-variables/)

https://user-images.githubusercontent.com/75349763/152619024-624abb4c-798e-4c9c-87fc-1e9a91ea8969.mp4

## Type of change

# How Has This Been Tested?

- [x] Create a .env file in frontend folder and add "REACT_APP_FEEDBACK_BUTTON=true" (or false) to toggle the button.

# Checklist:


aryanshah701 (commented):
LGTM! Where is the READ_APP_FEEDBACK_BUTTON environment variable coming from though? Did you export it on your local machine? If so then we'll need to configure the app for environment variables using `dotenv` and different files for each environment. Let's discuss this before merging.

.gitignore
```diff
@@ -56,6 +56,7 @@ typings/
 
 # dotenv environment variables file
 .env
+!frontend/.env*
```
aryanshah701, line 59: could you add a comment here to make it clear why we're excluding the frontend/.env? Something like "frontend env files shouldn't contain secrets" should be good!

frontend/.env
```diff
@@ -0,0 +1 @@
+REACT_APP_FEEDBACK_BUTTON=false
```
aryanshah701, line 1: let's make this `.env.development` and let's also add `env.production` so that we have separate environment files for the separate environments.

This will eventually be useful when we want some features in development but not production or vice versa. For now they will look the same since we don't want the feedback button in either!

aryanshah701 (approved):
Looks good! Just 2 more small changes. I'm approving it so that you can merge once you make those!

PR #384, opened 2022-02-10
fix completed courses bug

justineluo:
(if gif is not working, please ctrl+click to open in new tab)
Graduate PR #1 gif

# Description

Fixes a bug where transfer courses added during onboarding weren't correctly being added to the schedule as transfer courses. More specifically, fixes:

1. slowdown due to consecutive awaits for fetching courses (fix is to use Promise.all)
2. incorrect course equality checks
3. double-equals equality checks

Additionally, this cleans up some missing-key warnings, deduplicates some code,
and also adds text for the steps displayed during onboarding.

## Type of change

# How Has This Been Tested?

Tested locally by selecting completed courses in the onboarding workflow and checking that they show up in transfer credit holder.

I have not written any tests for my code.

# Checklist:

# Outstanding Questions

1. Should we change the strategy used to add completed courses to the schedule? (topological sort)
2. Is the text for steps displayed in the onboarding flow okay to add?

### High-level summary of changes:

- refactor TransferCoursesScreen to be functional component
- pull out shared utils from transfer screen and completed courses screen to course-helpers
- add explicitly controlled SelectableCourse component to GenericOnboarding

### DIFF
--- a/frontend/src/Onboarding/CompletedCoursesScreen.tsx
+++ b/frontend/src/Onboarding/CompletedCoursesScreen.tsx
@@ -15,7 +15,12 @@ import { Collapse, Grid, Link as ButtonLink, Paper } from "@material-ui/core";
 import { setCompletedRequirementsAction } from "../state/actions/studentActions";
 import { getUserMajorFromState } from "../state";
 import { AddClassSearchModal } from "../components/AddClassSearchModal";
-import { courseEq, coursesToString, flatten } from "../utils/course-helpers";
+import {
+  courseEq,
+  coursesToString,
+  courseToString,
+  flatten,
+} from "../utils/course-helpers";
 
 interface CompletedCoursesScreenProps {
   major: Major;
@@ -96,9 +101,7 @@ class CompletedCoursesComponent extends Component<Props, State> {
       <SelectableCourse
         key={coursesToString(courses)}
         checked={filtered.length !== requirements.length}
-        courseText={courses
-          .map(({ subject, classId }) => subject + classId)
-          .join(" and ")}
+        courseText={courses.map(courseToString).join(" and ")}
         onChange={e => {
           this.setState({
             completedRequirements: e.target.checked
--- a/frontend/src/Onboarding/CompletedCoursesScreen.tsx
+++ b/frontend/src/Onboarding/CompletedCoursesScreen.tsx
@@ -2,13 +2,8 @@ import React, { Component } from "react";
 import { connect } from "react-redux";
 import { AppState } from "../state/reducers/state";
 import { Dispatch } from "redux";
-import {
-  ScheduleCourse,
-  Major,
-  IRequiredCourse,
-  Requirement,
-} from "../../../common/types";
-import { withRouter, RouteComponentProps } from "react-router-dom";
+import { IRequiredCourse, Major, ScheduleCourse } from "../../../common/types";
+import { RouteComponentProps, withRouter } from "react-router-dom";
 import {
   CourseText,
   OnboardingSelectionTemplate,
@@ -16,34 +11,11 @@ import {
   TitleText,
 } from "./GenericOnboarding";
 import { AddBlock } from "../components/ClassBlocks/AddBlock";
-import { Link as ButtonLink, Collapse, Grid, Paper } from "@material-ui/core";
+import { Collapse, Grid, Link as ButtonLink, Paper } from "@material-ui/core";
 import { setCompletedRequirementsAction } from "../state/actions/studentActions";
 import { getUserMajorFromState } from "../state";
 import { AddClassSearchModal } from "../components/AddClassSearchModal";
-
-/**
- * Flattens the Requirement[] into only a list of Requirements/Requirement sets
- * This means that all inner lists will only contain one class or a list of the primary class and its labs/recitations
- * @param reqs
- */
-function flatten(reqs: Requirement[]): IRequiredCourse[][] {
-  return reqs.map(flattenOne).reduce((array, cur) => array.concat(cur), []);
-}
-
-function flattenOne(req: Requirement): IRequiredCourse[][] {
-  if (req.type === "COURSE") {
-    return [[req as IRequiredCourse]];
-  } else if (
-    req.type === "AND" &&
-    req.courses.filter(c => c.type === "COURSE").length
-  ) {
-    return [req.courses as IRequiredCourse[]];
-  } else if (req.type === "AND" || req.type === "OR") {
-    return flatten(req.courses);
-  } else {
-    return [];
-  }
-}
+import { flatten } from "../utils/course-helpers";
 
 interface CompletedCoursesScreenProps {
   major: Major;
--- a/frontend/src/Onboarding/CompletedCoursesScreen.tsx
+++ b/frontend/src/Onboarding/CompletedCoursesScreen.tsx
@@ -5,9 +5,9 @@ import { Dispatch } from "redux";
 import { IRequiredCourse, Major, ScheduleCourse } from "../../../common/types";
 import { RouteComponentProps, withRouter } from "react-router-dom";
 import {
+  SelectableCourse,
   CourseText,
   OnboardingSelectionTemplate,
-  SelectableCourse,
   TitleText,
 } from "./GenericOnboarding";
 import { AddBlock } from "../components/ClassBlocks/AddBlock";
@@ -15,7 +15,7 @@ import { Collapse, Grid, Link as ButtonLink, Paper } from "@material-ui/core";
 import { setCompletedRequirementsAction } from "../state/actions/studentActions";
 import { getUserMajorFromState } from "../state";
 import { AddClassSearchModal } from "../components/AddClassSearchModal";
-import { flatten } from "../utils/course-helpers";
+import { courseEq, coursesToString, flatten } from "../utils/course-helpers";
 
 interface CompletedCoursesScreenProps {
   major: Major;
@@ -61,29 +61,6 @@ class CompletedCoursesComponent extends Component<Props, State> {
     this.props.setCompletedRequirements(this.state.completedRequirements);
   }
 
-  /**
-   * Handles a class when it has been checked off. If it is being unchecked, it removes it,
-   * and if it is being checked, it adds it to the list of completed requirements
-   * @param e
-   * @param course
-   */
-  async onChecked(e: any, course: IRequiredCourse) {
-    const checked = e.target.checked;
-
-    if (checked) {
-      this.setState(prevState => ({
-        completedRequirements: [...prevState.completedRequirements, course],
-      }));
-    } else {
-      let courses = this.state.completedRequirements.filter(
-        c => c.subject !== course.subject && c.classId !== course.classId
-      );
-      this.setState({
-        completedRequirements: courses,
-      });
-    }
-  }
-
   // changes the expanding state of the specific section
   onExpand(requirementGroup: string, change: Boolean) {
     this.state.expandedSections.set(requirementGroup, change);
@@ -111,15 +88,24 @@ class CompletedCoursesComponent extends Component<Props, State> {
 
   // Renders one course/courseset (if it contains labs/recitiations, and seperated)
   renderCourse(courses: IRequiredCourse[]) {
-    let allCourse = courses.map(course => course.subject + course.classId);
-    const onChange = (e: React.ChangeEvent<HTMLInputElement>) =>
-      courses.forEach(course => this.onChecked(e, course));
-
+    const requirements = this.state.completedRequirements;
+    const filtered = requirements.filter(
+      r => !courses.some(c => courseEq(r, c))
+    );
     return (
       <SelectableCourse
-        key={allCourse[0]}
-        onChange={onChange}
-        courseText={allCourse.join(" and ")}
+        key={coursesToString(courses)}
+        checked={filtered.length !== requirements.length}
+        courseText={courses
+          .map(({ subject, classId }) => subject + classId)
+          .join(" and ")}
+        onChange={e => {
+          this.setState({
+            completedRequirements: e.target.checked
+              ? [...requirements, ...courses]
+              : filtered,
+          });
+        }}
       />
     );
   }
--- a/frontend/src/Onboarding/GenericOnboarding.tsx
+++ b/frontend/src/Onboarding/GenericOnboarding.tsx
@@ -24,7 +24,7 @@ const TitleLocation = styled.div`
   margin-top: 96px;
   margin-bottom: 48px;
   width: 256px;
-  font-family: .Helvetica Neue DeskInterface;
+  font-family: .Helvetica Neue DeskInterface, serif;
   font-style: normal;
   font-weight: bold;
   font-size: 24px;
@@ -125,6 +125,7 @@ const steps = [
 interface SelectableCourseProps {
   readonly onChange: (event: React.ChangeEvent<HTMLInputElement>) => void;
   readonly courseText: string;
+  readonly checked: boolean;
 }
 
 interface GenericOnboardingTemplateProps {
@@ -142,25 +143,8 @@ interface OnboardingSelectionTemplateProps {
 const SelectableCourse: React.FC<SelectableCourseProps> = ({
   courseText,
   onChange,
+  checked,
 }) => {
-  return (
-    <CourseWrapper>
-      <Checkbox
-        style={{ width: 2, height: 2 }}
-        icon={<CheckBoxOutlineBlankIcon style={{ fontSize: 20 }} />}
-        checkedIcon={
-          <CheckBoxIcon style={{ fontSize: 20, color: "#EB5757" }} />
-        }
-        onChange={onChange}
-      />
-      <CourseText>{courseText}</CourseText>
-    </CourseWrapper>
-  );
-};
-
-const ControlledSelectableCourse: React.FC<SelectableCourseProps & {
-  checked: boolean;
-}> = ({ courseText, onChange, checked }) => {
   return (
     <CourseWrapper>
       <Checkbox
@@ -205,7 +189,7 @@ const GenericOnboardingTemplate: React.FC<GenericOnboardingTemplateProps> = ({
 };
 
 const OnboardingSelectionTemplateComponent: React.FC<OnboardingSelectionTemplateProps &
-  RouteComponentProps<{}>> = ({
+  RouteComponentProps> = ({
   screen,
   mainTitleText,
   onSubmit,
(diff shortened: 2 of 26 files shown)


frontend/src/Onboarding/CompletedCoursesScreen.tsx
```diff
@@ -2,48 +2,20 @@ import React, { Component } from "react";
...
-    req.type === "AND" &&
-    req.courses.filter(c => c.type === "COURSE").length
-  ) {
-    return [req.courses as IRequiredCourse[]];
-  } else if (req.type === "AND" || req.type === "OR") {
-    return flatten(req.courses);
-  } else {
-    return [];
-  }
-}
```
bugsalexander, line 46: these were duplicated in the transfer courses screen, so i moved them to `course-helpers.ts`

frontend/src/Onboarding/GenericOnboarding.tsx
```diff
@@ -41,7 +41,6 @@ const Box = styled.div`
 
 const DotWrapper = styled.div`
   display: flex;
-  width: 256;
```
bugsalexander, line 44: removed static width on this so that the steps center properly

frontend/src/Onboarding/GenericOnboarding.tsx
```diff
@@ -113,7 +115,12 @@ const theme = createMuiTheme({
   },
 });
 
-const steps = ["", "", "", ""];
+const steps = [
+  "Student Profile",
+  "Completed Courses",
+  "Transfer Courses",
+  "AP/IB Courses",
+];
```
bugsalexander, line 123: added steps so that users can view what steps they have completed/remaining. we can totally change the text here. changed this due to a warning, see below

frontend/src/Onboarding/GenericOnboarding.tsx
```diff
@@ -151,6 +158,25 @@ const SelectableCourse: React.FC<SelectableCourseProps> = ({
   );
 };
 
+const ControlledSelectableCourse: React.FC<SelectableCourseProps & {
```
bugsalexander, line 161: The above `SelectableCourse` tries to be an [uncontrolled component](https://reactjs.org/docs/uncontrolled-components.html), but according to the MUI docs, [controlled components take an `onChange`](https://mui.com/components/checkboxes/#controlled).

I had tried using `SelectableCourse` as it had been used before, but for whatever reason I couldn't get it to be uncontrolled (clicking checkbox would not check the component). I'm guessing it's because we pass it an `onChange`? Though I'm not sure, since that seems to be elsewhere... it may also be context dependent (i'm using it in a function component, the others are classes).

TLDR was having trouble getting uncontrolled to work, so I made a controlled variant of `SelectableCourse`. 

If reviewers think it's cleaner, I can also add a `controlled` prop to the existing `SelectableCourse` component so that there's less code duplication.

frontend/src/Onboarding/TransferCoursesScreen.tsx
```diff
@@ -35,253 +30,159 @@ const TitleText = styled.div`
...
-    return [req.courses as IRequiredCourse[]];
-  } else if (req.type === "AND" || req.type === "OR") {
-    return flatten(req.courses);
-  } else {
-    return [];
-  }
-}
-
-interface TransferCoursesScreenProps {
+type TransferCoursesScreenProps = {
```
bugsalexander, line 33: Since I was refactoring anyway, I converted this to be a functional component as well, lmk thoughts on this

frontend/src/Onboarding/TransferCoursesScreen.tsx
```diff
@@ -35,253 +30,159 @@ const TitleText = styled.div`
...
+const TransferCoursesScreenComponent: React.FC<TransferCoursesScreenProps> = () => {
+  const major = useSelector((state: AppState) => getUserMajorFromState(state)!);
+  const completedRequirements = useSelector((state: AppState) =>
+    getCompletedRequirementsFromState(state)
+  );
+  const dispatch = useDispatch();
+  const setCompletedCourses = (completedCourses: ScheduleCourse[]) =>
+    dispatch(setCompletedCoursesAction(completedCourses));
+  const setTransferCourses = (transferCourses: ScheduleCourse[]) =>
+    dispatch(setTransferCoursesAction(transferCourses));
```
bugsalexander, line 62: first time using redux 🤪 so lmk if this looks ok or not

frontend/src/Onboarding/TransferCoursesScreen.tsx
```diff
@@ -35,253 +30,159 @@ const TitleText = styled.div`
...
    * Separate transfer and non transfer courses, convert non transfer completed courses
    * into ScheduleCourses, and populate redux store.
    */
-  async onSubmit() {
-    let completedNonTransfer = this.state.completedNonTransfer.filter(
-      req => !this.state.selectedRequirements.includes(req)
+  const onSubmit = async () => {
+    const filteredCompleted = completedNonTransfer.filter(
+      req => !selectedRequirements.some(c => courseEq(c, req))
     );
```
bugsalexander, line 120: this is what I think was the first bug: `includes` uses ref equality (bad). it may have been fine, but it's generally an antipattern to use ref equality.

frontend/src/Onboarding/TransferCoursesScreen.tsx
```diff
@@ -35,253 +30,159 @@ const TitleText = styled.div`
...
-    let completed: ScheduleCourse[] = [];
-
-    for (let course of completedNonTransfer) {
-      const scheduleCourse = await fetchCourse(
-        course.subject,
-        String(course.classId)
-      );
```
bugsalexander, line 127: (semi) bug # 2 `await` inside of for-loop (bad)

frontend/src/Onboarding/TransferCoursesScreen.tsx
```diff
@@ -35,253 +30,159 @@ const TitleText = styled.div`
...
-   * and if it is being checked, it converts it to a ScheduleCourse and adds it to selected courses
-   * @param e
-   * @param course
-   */
-  async onChecked(e: any, course: IRequiredCourse) {
-    const checked = e.target.checked;
-
-    if (checked) {
-      this.setState({
-        selectedRequirements: [...this.state.selectedRequirements, course],
```
bugsalexander, line 161: bug # 3: this was used to set state multiple times for a single checkbox (ie checking CS2510 and CS2511), and so one of them was getting wiped out (only setting the second of the two, CS2511, as checked).

frontend/src/Onboarding/TransferCoursesScreen.tsx
```diff
@@ -35,253 +30,159 @@ const TitleText = styled.div`
...
-      let reqs = this.state.selectedRequirements.filter(r => r !== course);
-      this.setState({
-        selectedRequirements: reqs,
-      });
-    }
-  }
+    const [completed, transfer] = await Promise.all([
+      fetchAndFilterCourses(filteredCompleted),
+      fetchAndFilterCourses(selectedRequirements),
+    ]);
```
bugsalexander, line 93: fix for await bug is to use promise.all

frontend/src/Onboarding/TransferCoursesScreen.tsx
```diff
@@ -35,253 +30,159 @@ const TitleText = styled.div`
...
-  /**
-   * Renders one course/courseset (if it contains labs/recitiations, and seperated)
-   * @param courses - Course pairings to be rendered
-   */
-  renderCourse(courses: IRequiredCourse[]) {
-    let allCourse = courses.map(course => course.subject + course.classId);
-    return (
-      <SelectableCourse
-        courseText={allCourse.join(" and ")}
-        onChange={e => courses.forEach(course => this.onChecked(e, course))}
```
bugsalexander, line 231: bug # 3 part 2: really this should be _one_ state update for all the courses in the set, not several updates (one per course). the later setState calls will override the beginning ones

frontend/src/Onboarding/TransferCoursesScreen.tsx
```diff
@@ -35,253 +30,159 @@ const TitleText = styled.div`
...
-      )
-      .map((r: IRequiredCourse[]) => this.renderCourse(r));
-  }
+  const CourseSet: React.FC<{ courses: IRequiredCourse[] }> = ({ courses }) => {
+    const filtered = selectedRequirements.filter(
+      r => !courses.some(c => courseEq(r, c))
+    );
+    return (
+      <ControlledSelectableCourse
+        checked={filtered.length !== selectedRequirements.length}
```
bugsalexander, line 108: pass in `checked`, so this is a controlled component

frontend/src/utils/completed-courses-helpers.ts
```diff
@@ -1,32 +1,11 @@
...
   ScheduleYear,
-  SeasonWord,
 } from "../../../common/types";
 import { convertToDNDSchedule } from "./schedule-helpers";
 
-/**
- * Returns the sum of all credits in the courses
- * @param courses
- */
-export function sumCreditsFromList(courses: DNDScheduleCourse[]): number {
```
bugsalexander, line 18: this was unused, so i deleted it. lmk if i should add it back

frontend/src/utils/completed-courses-helpers.ts
```diff
@@ -84,83 +69,62 @@ export function getNextTerm(is_summer: boolean, classes: DNDScheduleCourse[]) {
   return classes.splice(0, counter);
 }
 
-export function getSimplifiedCourseData(
```
bugsalexander, line 87: unused pt2

frontend/src/utils/completed-courses-helpers.ts
```diff
@@ -84,83 +69,62 @@ export function getNextTerm(is_summer: boolean, classes: DNDScheduleCourse[]) {
...
-  const termMap: { [key: string]: SeasonWord } = {
-    "10": "fall",
-    "30": "spring",
-    "40": "summer1",
-    "60": "summer2",
+
+  addBlankScheduleYear(years, yearMap, 2022);
+  yearMap[years[0]].fall.classes = [...completedCourses];
+  const [schedule, counter] = convertToDNDSchedule({ years, yearMap }, 0);
+  return { schedule, counter };
```
bugsalexander, line 86: Question: for this, we can actually do better. my idea is to use topological sort (implemented in json_parser) to create an ordering of courses in prereq-order. thoughts? otherwise for now, we just throw all the completed courses in fall and let the user handle moving them around.

scrapers/src/json_converter.ts
```diff
@@ -34,6 +34,8 @@ export const parseRequirement = (
       return parseRequiredRange(req);
     case "COURSE":
       return parseRequiredSimple(req);
+    case "CREDITS":
+      throw new Error("unimplemented!");
```
bugsalexander, line 38: just to make `yarn tsc` work

bugsalexander:
[Debug why transfer courses from onboarding are partially displayed in transfer holder](https://trello.com/c/18KZppk6/382-debug-why-transfer-courses-from-onboarding-are-partially-displayed-in-transfer-holder)

frontend/src/Onboarding/GenericOnboarding.tsx
aryanshah701: interesting, why do you think we are doing uncontrolled select components for some areas then(completed course screen for example)? if react suggests controlled components what makes our case unique in requiring an uncontrolled select component?

aryanshah701: What do you think about abstracting the component into a general select with the `isControlled` prop but also exporting `UncontrolledSelectableCourse` and `ControlledSelectableCourse` that essentially just wrap around the generic select and pass down true or false for the `isControlled` prop. 

If you find that clean that would be awesome! Otherwise just adding the prop is fine but lets maybe abstract out the common MUI styling and stuff :)

frontend/src/Onboarding/TransferCoursesScreen.tsx
aryanshah701: awesome!!!

frontend/src/Onboarding/TransferCoursesScreen.tsx
aryanshah701: good catch!!

frontend/src/Onboarding/TransferCoursesScreen.tsx
aryanshah701: ooh I like `useDispatch` and `useSelector`! Curious are there any benefits to using these as opposed to `connect` or is this just the more functional/hooks way of doing things?

frontend/src/Onboarding/TransferCoursesScreen.tsx
aryanshah701: sweet!

frontend/src/utils/completed-courses-helpers.ts
aryanshah701: i think that's an awesome idea, it doesn't seem too bad to implement too! bring this up during standup and @justineluo can add it to our backlog and prioritize it!

frontend/src/utils/completed-courses-helpers.ts
```diff
@@ -84,83 +69,62 @@ export function getNextTerm(is_summer: boolean, classes: DNDScheduleCourse[]) {
...
+): { schedule: DNDSchedule; counter: number } {
   const years: number[] = [];
   const yearMap: { [key: number]: ScheduleYear } = {};
-  const termMap: { [key: string]: SeasonWord } = {
-    "10": "fall",
-    "30": "spring",
-    "40": "summer1",
-    "60": "summer2",
+
+  addBlankScheduleYear(years, yearMap, 2022);
```
aryanshah701, line 83: question: why do we have only add a 2022 blank schedule year? And why specifically 2022? Do we just dump completed courses into a 2022 year or am i misunderstanding this?

scrapers/src/json_converter.ts
aryanshah701: woah, did yarn tsc not work before or is this related to some change made in this pr??

frontend/src/utils/course-helpers.ts
```diff
@@ -55,3 +57,32 @@ export async function getScheduleCourseCoreqs(
...
+    return flatten(req.courses);
+  } else {
+    return [];
+  }
+}
+
+export const courseToString = (c: {
+  subject: string;
+  classId: number | string;
+}) => `${c.subject}${c.classId}`;
```
aryanshah701, line 88: Thank you!!! I think we have course to string being done in so many places but nobody bothered to standardize it. If you have time and saw some of it would really appreciate it if you could add this function but don't spend more than a few minutes!(hint: line 114 in CompletedCoursesScreen)

aryanshah701 (commented):
Amazing stuff!! Loved the adhoc changes to clean things up to whichever files you touched <3. Just have a couple of questions regarding the changes + 1 small refactor request of the Selectable component, otherwise all good to merge!

frontend/src/Onboarding/LoginScreen.tsx
```diff
@@ -58,11 +58,6 @@ const ButtonContainer = styled.div`
 `;
 
 interface ReduxStoreLoginScreenProps {
-  setAcademicYear: (academicYear: number) => void;
-  setGraduationYear: (graduationYear: number) => void;
-  setUserCoopCycle: (coopCycle: string) => void;
-  setSchedules: (schedules: NamedSchedule[]) => void;
-  setPlanStr: (planStr: string) => void;
```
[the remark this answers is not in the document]
bugsalexander: shoutout to @justineluo

bugsalexander:
> interesting, why do you think we are doing uncontrolled select components for some areas then(completed course screen for example)? if react suggests controlled components what makes our case unique in requiring an uncontrolled select component?

not sure. my best guess is that at the time, it was easier to do, or that we weren't aware of doing so.

bugsalexander:
> What do you think about abstracting the component into a general select with the isControlled prop but also exporting UncontrolledSelectableCourse and ControlledSelectableCourse that essentially just wrap around the generic select and pass down true or false for the isControlled prop.

yep! i may also just replace all the instances of the uncontrolled variant, if it's easy enough to do.

frontend/src/Onboarding/TransferCoursesScreen.tsx
bugsalexander: yes, this is just the functional way of doing things. and there aren't "benefits" per se, but the API is slightly different/more flexible, see https://react-redux.js.org/api/hooks#useselector

frontend/src/Onboarding/GenericOnboarding.tsx
bugsalexander: not sure. my best guess is that at the time, it was easier to do, or that we weren't aware of doing so.

and yep! i may also just replace all the instances of the uncontrolled variant, if it's easy enough to do.

scrapers/src/json_converter.ts
bugsalexander: didn't work before

frontend/src/utils/completed-courses-helpers.ts
bugsalexander: oh yeah i need to fix this, thanks!

frontend/src/utils/course-helpers.ts
bugsalexander: i'll do a look over!

frontend/src/Onboarding/GenericOnboarding.tsx
bugsalexander: test comment, ignore

frontend/src/Onboarding/GenericOnboarding.tsx
aryanshah701: sweet thanks!!

scrapers/src/json_converter.ts
aryanshah701: crazy

frontend/src/utils/completed-courses-helpers.ts
bugsalexander: just tested it out and it seems like currently the years get overwritten to start at graduating year - current years at neu, and dont' actually use this at all. i'll change it to a constant and leave a comment

bugsalexander: previously it was getting overwritten in `studentReducer.ts` by `alterScheduleToHaveCorrectYears` but rather than having two tightly-coupled functions, i just pulled out a function. should be good now

frontend/src/utils/course-helpers.ts
bugsalexander: done

aryanshah701 (approved):
looks good to me!! also thank you so much for the courseToString changes!

PR #392, opened 2022-03-02
Add year button

jia614:
add-year-button

# Description

Added spacing between the years, removed spacing between semesters.

Created an "Add Year +" button that adds a new year to the user's current plan. On plan update, button is disabled.

## Type of change

# How Has This Been Tested?

- Click add year, then look at state to ensure that year has been added to plan. Should be reflected in the home page as a new year. 

# Checklist:


frontend/src/components/Schedule/AddYearButton.tsx
```diff
@@ -0,0 +1,91 @@
...
+      }}
+      disabled={isUpdating}
+      onClick={handleNewSemester}
+    >
+      + Add Year
+    </WhiteColorButton>
+  );
+};
+
+const mapStateToProps = (state: AppState) => ({
```
bugsalexander, line 71: Rather than use mapStateToProps and mapDispatchToProps, use the redux hooks api. See here for [reference](https://github.com/sandboxnu/graduatenu/blob/main/frontend/src/Onboarding/TransferCoursesScreen.tsx).

It's easier to read and also is more similar to the react hooks api, and [is the recommended api to use by redux docs](https://react-redux.js.org/api/hooks)-see tip at top of page.

frontend/src/services/PlanService.ts
```diff
@@ -218,6 +219,65 @@ export const fetchComments = (planId: number, studentId: number) =>
...
+        year: currYear,
+        termId: currYear * 100 + 60,
+        status: "INACTIVE",
+        classes: [],
+      },
+    },
+  };
+  const newPlan = { ...plan, schedule: { ...schedule, years, yearMap } };
+
+  return fetch(`/api/users/${userId}/plans/${planId}`, {
```
bugsalexander, line 271: Do we have axios as a dependency? If so, use that over fetch.

bugsalexander (commented):
I can help pair to make changes later today!

frontend/src/components/Schedule/AddYearButton.tsx
cindy1u0: I think we have been using `mapStateToProps` and `mapDispatchToProps`. If we want, we should prob start using those hooks from now on 8)

bugsalexander: Yeet sounds good, i want

cindy1u0: LMAOOO "i want"

bugsalexander: `y e s`

frontend/src/services/PlanService.ts
jia614: we don't have axios as a dependency :(

bugsalexander (approved):
poggers! ty for pairing ❤️

PR #394, opened 2022-03-09
Setup running new backend

aryanshah701:
# Description

Please include a summary of the change and which issue is fixed.
Please also include relevant motivation and context. 
List any dependencies that are required for this change.

Closes # (issue number)

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

# Checklist:


.github/workflows/main.yml
```diff
@@ -5,17 +5,13 @@ on: [push]
...
-        node: ["12", "13"]
-      fail-fast: false
     name: Scraper Tests - Node ${{ matrix.node }} compatibility
     steps:
       - uses: actions/checkout@v1
       - name: Setup node
         uses: actions/setup-node@v1
         with:
-          node-version: ${{ matrix.node }}
+          node-version: "16"
```
aryanshah701, line 14: v13 compatibility was failing cause the `concurrently` package that this PR adds isn't compatible with v13. Thought It's probably best to go up to v16 anyways especially since our code works on v16?

cindy1u0: tbh I think node v16 has a lot of issues... not sure if it's very feasible to change that to v16 maybe @bugsalexander might have more thoughts on this

frontend/.env.development
```diff
@@ -1 +1,3 @@
-REACT_APP_FEEDBACK_BUTTON=false
\ No newline at end of file
+REACT_APP_FEEDBACK_BUTTON=false
+PORT=3000
+BROWSER=none
```
aryanshah701, line 3: `start:v2` run graduate on port 3002, not on 3000. BROWSER=none stops the browser from opening up when the frontend runs. Did this since it is confusing to have the browser open at port 3000 when the app actually runs on 3002.

.github/workflows/main.yml
aryanshah701: oh I didn't know that, thought it was fine since our tests pass and the current LTS version of node is v16

cindy1u0: hmm I guess but I am also just thinking whether or not v16 might cause problems for some incompatibility with some packages we have in the codebase

aryanshah701: Yeap that makes sense but this is also just the node version for CI running our scraper tests(which work). Also if packages are incompatible with v16 then the `yarn install` would complain. And from some quick googling node actually hasn't had breaking changes for the past 2 major releases. Oh and we're kinda redoing the scrapper anyways right lmao?

cindy1u0: lol that's true

.github/workflows/main.yml
```diff
@@ -5,17 +5,13 @@ on: [push]
 jobs:
   run-scrapers-tests:
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        node: ["12", "13"]
-      fail-fast: false
     name: Scraper Tests - Node ${{ matrix.node }} compatibility
```
bugsalexander, line 8:
```suggestion
    name: Scraper Tests
```

.github/workflows/main.yml
bugsalexander: i'd probably stick with 14 for now since it's battle tested, we can always bump version to 16 later?

package.json
```diff
@@ -12,6 +12,10 @@
   ],
   "scripts": {
     "start": "cd infrastructure/develop && ./compose-build.sh && ./run-servers.sh dev-build",
+    "start:v2": "concurrently \"yarn dev:server:v2\" \"yarn dev:app\" \"yarn dev:proxy\"",
+    "dev:server:v2": "cd api-v2 && yarn dev",
+    "dev:app": "cd frontend && yarn start",
```
bugsalexander, line 17: later on when we switch to yarn workspaces, can do `yarn workspace <workspace> <cmd>` see [here](https://classic.yarnpkg.com/en/docs/cli/workspace)

bugsalexander (approved):
fix changes, then good to ship

.github/workflows/main.yml
bugsalexander: we should probably use the same node version for everything (frontend, scrapers, backend) so i'd say use 14 for now since we don't know if all will work on 16.

package.json
aryanshah701: yeap already have the change in the workspace branch!

PR #405, opened 2022-03-27
Migrate to Yarn 2 + Common Package Build Fix

aryanshah701:
# Description

The frontend production build is broken in the browser because the common package isn't built properly for the browser. The common package needs to be built in different ways depending on who's using it. For instance, a nodejs server using the common package needs to have the package built differently than a frontend application running in the browser.

Details from slack thread on solution

Essentially we can handle common packages in 2 ways:
- Build them separately and then use the builds in apps
- Delegate the building to the apps themselves

Option 1 requires build configuration upfront but then we never have to worry about configuring webpack every time for new apps/services. This isn’t too bad though and KOH does this - both the Nest and Next apps have custom webpacks that build the common packages.

If we we are going for option 1 then we essentially need to configure babel to build for 3 main cases
- api-v2: Nodejs webpacked code
- frontend: Browser webpacked code
- dev: No webpacked code

A [GitHub issue](https://gist.github.com/ncochard/6cce17272a069fdb4ac92569d85508f4) that explains this in more detail.

After I made this fix, CI started failing for odd yarn related reasons. This may have to do with minor bugs in yarn, Alex found it easiest to "debug" this by migrating to yarn v2 which is AWESOME AND BLAZINGLY FAST.

*TLDR: Staging was broken cause of babel issues, fix to babel issues broke CI for arbitrary yarn reasons, Alex is chungus and migrated us to yarn V2 which is AWESOME*.

Closes # No issue was created :( 

## Type of change

# How Has This Been Tested?

- CI passes which means docker is able to build the images fine + tests run ok
- `yarn dev` to run the app locally
- `yarn workspace @graduate/frontend start` to run the production build of our frontend (if this works staging *should* be fixed lol)

# Checklist:


bugsalexander:
seems like this is a problem related to the npm-install action we're using: https://github.com/bahmutov/npm-install/issues/28 (scroll down to bottom) – it's supposed to cache node modules to speed up subsequent installs, but looks like it's not actually working/is causing an error.

just remove it for now, who cares if CI is a bit slower, better to have it work and be slow than not work at all

bugsalexander:
https://github.com/yarnpkg/yarn/issues/2629#issuecomment-753480348

YARN_MIGRATION.md
```diff
@@ -0,0 +1,23 @@
...
+
+NOTE: `yarn install --immutable` requires ALL package.jsons, otherwise it will error (see frontend dockerfile)
+
+to get frontend build to work, i just copied all the root devDeps to common package.json, but it doesn't need all of them, so remove the non-important ones.
+
+### re-configure shared build scripts and dependencies
+
+so to build, we use babel. that's fine (it can remain at top-level) but the way we use it needs to be updated.
+
+see https://yarnpkg.com/getting-started/qa#how-to-share-scripts-between-workspaces
```
aryanshah701, line 23: can we delete this?

packages/common/prereq_loader.ts
```diff
@@ -50,7 +50,7 @@ export async function addPrereqsToSchedule(
   const loader: DataLoader<SimpleCourse, PrereqQueryResult> = new DataLoader<
     SimpleCourse,
     PrereqQueryResult
-  >(queryCoursePrereqData);
+  >(queryCoursePrereqData as any);
```
aryanshah701, line 53: 👀

aryanshah701 (commented):
LGTM! we probably want somebody else to review though lol. 

OOH also, what do the yarn plugins(plugin-ts and workspace-tools) do? Seems like you had to make some TS related changes, were they related to this?

packages/common/package.json
```diff
@@ -2,13 +2,21 @@
...
+  "browser": "lib/index.js",
   "private": true,
   "scripts": {
-    "dev": "yarn build -w",
-    "build": "babel --root-mode upward -d dist -x '.ts' *.ts"
+    "dev": "yarn build:dev -w",
+    "build": "yarn build:dev && yarn build:browser && yarn build:module",
+    "build:dev": "yarn g:babel --root-mode upward -d dist -x '.ts' *.ts",
+    "build:browser": "yarn g:cross-env BABEL_ENV=browser yarn g:babel --root-mode upward -d lib -x '.ts' *.ts",
+    "build:module": "yarn g:cross-env BABEL_ENV=module yarn g:babel --root-mode upward -d module -x '.ts' *.ts"
```
aryanshah701, line 13: build the common package 3 different ways into dist/ browser/ module/. dist/ is used in dev, browser/ is used by frontend application and module/ is used by NodeJS applications(api-v2)

bugsalexander:
@aryanshah701 
> OOH also, what do the yarn plugins(plugin-ts and workspace-tools) do? Seems like you had to make some TS related changes, were they related to this?

[Here's the relevant docs](https://yarnpkg.com/features/plugins#gatsby-focus-wrapper) but basically the typescript one will auto add the corresponding "@types/pkg" if the pkg doesn't contain the typedefs, and then the worjspaces one is so we can do things like "yarn workspaces foreach <cmd>"

The TS changes I think were related to the typescript version changing when i added it to the packagejson of the child packages... i think

YARN_MIGRATION.md
```diff
@@ -0,0 +1,23 @@
+# YARN 2 MIGRATION DOC
+
+## Remaining todos:
+
+### fix all missing dependencies
+
+to see problems, run `yarn dlx @yarnpkg/doctor`. note: `yarn global` no longer exists. instead, yarn has the equivalent of `npx` (roughly) via `yarn dlx`.
```
cindy1u0, line 7: just tried this out - this feature is pretty cool! We have some errors ;-;-;-;

YARN_MIGRATION.md
```diff
@@ -0,0 +1,23 @@
+# YARN 2 MIGRATION DOC
```
cindy1u0, line 1: thinking that this file just contains all the TODOs right, should this be in notion instead of in the codebase? Wondering if we should use this migration doc to document why/how we changed to yarn2? No big deal otherwise just curious

package.json
```diff
@@ -8,20 +8,22 @@
...
   "scripts": {
-    "dev": "concurrently -n \"common,api,frontend\" \"yarn workspace @graduate/common dev\" \"yarn workspace @graduate/api dev\" \"yarn workspace @graduate/frontend dev\"",
-    "dev:v2": "concurrently -n \"common,api-v2,frontend,proxy\" \"yarn workspace @graduate/common dev\" \"yarn workspace @graduate/api-v2 dev\" \"yarn workspace @graduate/frontend dev\" \"yarn dev:proxy\"",
+    "dev": "concurrently -n \"common,api,frontend\" \"yarn packages/common dev\" \"yarn packages/api dev\" \"yarn packages/frontend dev\"",
+    "dev:v2": "concurrently -n \"common,api-v2,frontend,proxy\" \"yarn packages/common dev\" \"yarn packages/api-v2 dev\" \"yarn packages/frontend dev\" \"yarn dev:proxy\"",
     "dev:proxy": "node infrastructure/develop/dev-proxy.js",
     "dev:db:up": "docker-compose -f infrastructure/develop/docker-compose.yml up -d",
     "dev:db:down": "docker-compose -f infrastructure/develop/docker-compose.yml down",
-    "prod:start": "concurrently \"yarn workspace @graduate/api dev\" \"yarn workspace @graduate/frontend docker:start\"",
+    "prod:start": "concurrently \"yarn packages/api dev\" \"yarn packages/frontend docker:start\"",
```
cindy1u0, line 19: oh just curious though i haven't run it - why do we not need to run `yarn packages/common` here? Is it cuz that package is just for dev usage?

packages/common/prereq_loader.ts
```diff
@@ -74,7 +74,7 @@ export async function addPrereqsToSchedules(
...
```
cindy1u0, line 77: 👀

cindy1u0 (approved):
yall cracked

package.json
aryanshah701: good question! So currently common package is only being used by the frontend(the old backend doesn't use it). The `prod:start` command run the frontend in docker. The Dockerfile for the frontend is responsible for copying over the common directory along with the frontend and building them together!

Eventually this command will run the new backend with the frontend... if we decide to Dockerize our new backend then the Dockerfile for the backend will copy over the backend + common + any other dependency package and build them all. But this is far out haha.

YARN_MIGRATION.md
aryanshah701: Putting this stuff in a `Guide to Migrate to Yarn V2` in notion might be a good idea since it'll make it easier for other teams to do the same stuff

YARN_MIGRATION.md
aryanshah701: haha yea :(

package.json
cindy1u0: ah gotcha thankss

YARN_MIGRATION.md
cindy1u0: ya just not sure if this doc should go into the codebase

aryanshah701: yea 100%

YARN_MIGRATION.md
bugsalexander: yep LOL

packages/common/prereq_loader.ts
bugsalexander: dont worry about it HAHAHA

YARN_MIGRATION.md
bugsalexander: yep!

PR #408, opened 2022-04-11
Update all frontend dependencies

bugsalexander:
# Description

Right now, we are working on developing the `@graduate/common` package (via the api-client branch), adding things like our shared dto-types (which use decorators).

However, the frontend depends on the `@graduate/common` package, meaning the frontend needs to be able to process the build output of the common module. Previously it could not do this because the dependencies were too old.

This PR updates the dependencies in `@graduate/frontend` so that they can process the decorator build output of the common module.

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# How to test

1. run `yarn` to install new dependencies
2. run `yarn dev` to launch dev servers
3. go to create an account. proceed through the workflows like normal, selecting whatever you wish.
4. once on home, drag around some courses, etc. stuff should not break.
5. logout
6. login with the account you just created
7. once on home, can play around again

verify that nothing breaks when doing the above.

# Notes

- ran `yarn npm-check-updates -u` in the frontend dir to update all deps in package.json automatically
- i uninstalled dataloader because it contains a [breaking bug](https://github.com/graphql/dataloader/issues/249) for all non-internet explorer browsers
  - note: this has been fixed by [this PR](https://github.com/graphql/dataloader/pull/293), but it is unreleased
  - because of this, we may fetch the same course multiple times from search, but we can fix it in frontend v2
- ~~i haven't actually tested this with the new decorator changes in `@graduate/common` but i will ASAP~~ 
  - **tested: it fixes them! new branch is on `api-client-2`**


aryanshah701 (approved):
looks good to me!!

PR #410, opened 2022-04-12
Add toasts, logging, and error boundary

aryanshah701:
# Description

Basic error handling(toasts for api errors, error boundary for client side) and logging.

- Added a logger utility that simply wraps console.log functionality. As our application matures, the logger can be made more sophisticated.
- Added a toast utility to display toasts using react-toastify. Currently a simple wrapper over react-toastify but the logic will mature as needed + we can easily migrate to another library/update the toast logic across the application.
- Added an ErrorBoundary that renders a very basic error page for now. Once, material UI and styled components are configured, the error page will be styled.

Closes [Error Handling](https://trello.com/c/QFjxMXRJ/430-error-handling)

## Type of change

# How Has This Been Tested?

Buttons on the landing page to trigger all of the functionality added. Click on the buttons to view the toasts, logs in the console, or trigger a client side error that the Error Boundary catches

Note: `logger.debug` logs are not visible in the console by default. To view them you'll have to select the "verbose" option as shown below.

Screen Shot 2022-04-12 at 5 19 07 PM

# Checklist:

### DIFF
--- a/packages/frontend-v2/components/Error/ClientSideError.tsx
+++ b/packages/frontend-v2/components/Error/ClientSideError.tsx
@@ -1,0 +1,12 @@
+export const ClientSideError: React.FC = () => {
+  return (
+    <>
+      <div>
+        <h1>GraduateNU</h1>
+      </div>
+      <div>
+        <p>There was an error rendering your page.</p>
+      </div>
+    </>
+  );
+};
--- a/packages/frontend-v2/pages/index.tsx
+++ b/packages/frontend-v2/pages/index.tsx
@@ -1,4 +1,3 @@
-import "react-toastify/dist/ReactToastify.min.css";
 import type { NextPage } from "next";
 import { toast, logger } from "../utils";
 import { useState } from "react";
--- a/packages/frontend-v2/pages/index.tsx
+++ b/packages/frontend-v2/pages/index.tsx
@@ -1,22 +1,105 @@
+import "react-toastify/dist/ReactToastify.min.css";
 import type { NextPage } from "next";
-import { SeasonEnum } from "@graduate/common";
+import { toast, logger } from "../utils";
+import { useState } from "react";
+import { ClientSideError } from "../components/Error/ClientSideError";
 
 const Home: NextPage = () => {
-  // using a type from our common package
-  const seasons: SeasonEnum[] = [
-    SeasonEnum.FL,
-    SeasonEnum.SP,
-    SeasonEnum.S1,
-    SeasonEnum.S2,
-  ];
+  const [isClientSideError, setIsClientSideError] = useState(false);
+
+  if (isClientSideError) {
+    return <ClientSideError />;
+  }
 
   return (
     <>
       <h1>GraduateNU Landing Page:</h1>
-      <h2>Seasons typed using types from our common package!</h2>
-      {seasons.map((season) => (
-        <p key={season}>{season}</p>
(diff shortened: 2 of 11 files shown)


bugsalexander:
[Error Handling](https://trello.com/c/QFjxMXRJ/430-error-handling)

packages/frontend-v2/pages/index.tsx
```diff
@@ -1,22 +1,108 @@
+import "react-toastify/dist/ReactToastify.min.css";
```
cindy1u0, line 1: I also see this in the documentation. Do we have to do this in every file if we were to use toasts?

bugsalexander: also ^ should it be here, or in _app.tsx?

packages/frontend-v2/components/Error/ClientSideError.tsx
```diff
@@ -0,0 +1,12 @@
+export const ClientSideError: React.FC = () => {
+  return (
+    <>
+      <div>
```
bugsalexander, line 4: are we bringing over the styling/other details from the old error page? could also be done later

bugsalexander (approved):

packages/frontend-v2/utils/logger.ts
```diff
@@ -0,0 +1,16 @@
...
+/**
+ * A logger utilty that encapsulates all logging logic. This will eventually have logic that
+ * handles production logging using tools like sentry.
+ */
+const log = (type: LogType, ...messages: unknown[]) => {
+  console[type](...messages);
+};
+
+export const logger = {
+  info: (...messages: unknown[]) => log("info", ...messages),
```
cindy1u0, line 12: hmm maybe im trolling but why is this unknow[] instead of just string? Is it just how console works? I just thought the type unknown[] doesn't really provide that much info

packages/frontend-v2/components/Error/ClientSideError.tsx
aryanshah701: yeap! I forgot to put in a note but i plan to do that once the styling PR gets merged!

packages/frontend-v2/utils/logger.ts
aryanshah701: haha yeap type unknown[] doesn't provide any information but the idea is that we want the logger to by able to take a comma separated list of anything in(objects, strings, number, etc) just like the console functions do. 

I could type it as any[] but unknown[] is stricter since it prevents the logger from assuming anything about the data that's coming in. And frankly i think the logger shouldn't know/care about it is getting especially if it's just passing it to the console for now

packages/frontend-v2/pages/index.tsx
aryanshah701: good catch this should go in app.tsx! 

but yeap we'll need to add the styles to all our pages that use toasts... if we were using CRA then we wouldn't need to do this but with Nextjs UI libraries often find it easier to bundle their styles and have them imported since Next doesn't allow components styles to be injected in the document for any static/server side pages using `style-loader`. Dealt with this issue a ton on my co-op lmaoo.

packages/frontend-v2/utils/logger.ts
cindy1u0: gotcha

cindy1u0 (approved):

PR #411, opened 2022-04-19
Aryan/style and theme

aryanshah701:
# Description

- Add Chakra UI as a dependency
   - Context: Justine, Cindy and I spoke about using Chakra UI instead of Material/Styled components. There isn't anything explicitly wrong with MUI/Styled but Chakra seems to suit our needs more(apologies, this should have been done during the platform design review 😞). 
     - Chakra UI is a layout-focused library and has a much greater emphasis on flexible, composable, and scalable code. The primary difference between it and MUI is 'Ease of Modification'. MUI is great to boostrap projects quickly and move fast, but given how our app has completely customized figma designs, Chakra makes it much easier to implement those designs with the library and without having to write any CSS/use styled components. 
     - For example, we have so many styled components in our component files. The reason for this is it is much easier to style a components from scratch than modify and existing MUI component. The leads to 2 different patterns of styling where we use MUI for large complex components, and styled components for layout, and simple components. This is totally fine but if we are using a UI Library, why not use one that allows us to do everything we want. 
     - If anyone wants to talk and open up this discussion reach out! I'm fully open to hearing your thoughts!
 
- Theming using Cindy's design doc for responsive breakpoints, and colors from our [design system](https://www.figma.com/file/CPqFE7T1aJe46zR99n0SPr/GraduateNU-Design-System?node-id=2%3A5). The theme will mature as we start collaborating with our designer to establish our concrete design system. For now, it's there so that when modifications happen to things like our primary color, spacing, etc. it's reflected throughout our app easily.

Note: The [styling and design platform doc](https://www.notion.so/sandboxnu/Styling-and-Theming-8593a27165d646dab73409274c4af996) needs to be updated to reflect any changes in our decisions. I have currently just added a [Chakra UI section](https://www.notion.so/sandboxnu/Styling-and-Theming-8593a27165d646dab73409274c4af996#1a4b49abc14f4d5ab45d51833a07e8c1) with some basic best practices. Once this PR is approved and all decisions are final, I'll go and update the doc.

Closes: [Styling and Theming](https://trello.com/c/tNSb1yop/434-styling-and-theming)

## Type of change

# How Has This Been Tested?

Use chakra UI in the landing page. Rin `yarn dev` from the frontend-v2 folder to checkout the results.

# Checklist:


packages/frontend-v2/pages/index.tsx
```diff
@@ -1,23 +1,13 @@
...
-
   return (
-    <>
-      <h1>GraduateNU Landing Page:</h1>
-      <h2>Seasons typed using types from our common package!</h2>
-      {seasons.map((season) => (
-        <p key={season}>{season}</p>
-      ))}
-    </>
+    <Flex width="full" justifyContent="center" my="xl">
```
aryanshah701, line 6: seeee we can use things like "<Flex />, and "<Box />" without having to create our styled.divs. These components by default are better than styled.divs cause they have easy access to theme values(spacing, colors, responsiveness). Chakra has a whole suite of layout components too 👀

packages/frontend-v2/pages/index.tsx
```diff
@@ -1,23 +1,13 @@
...
+      <Heading color="primary.main" size="xl">
```
aryanshah701, line 7: size="xl" makes the Heading automatically responsive(size decreases on smaller screen sizes, and vice-versa) in a way that is consistent with the entire application(all components follow the same responsive guidelines).

packages/frontend-v2/utils/theme.ts
```diff
@@ -0,0 +1,83 @@
+import { extendTheme } from "@chakra-ui/react";
```
aryanshah701, line 1: Theme values are currently just whatever i could get from our design system and cindy's notion(for breakpoints). This will evolve eventually.

cindy1u0: are those colors from the old design system or the new one that linda has?

packages/frontend-v2/package.json
```diff
@@ -9,7 +9,11 @@
     "lint": "next lint"
   },
   "dependencies": {
+    "@chakra-ui/react": "^1.8.8",
+    "@emotion/react": "^11",
+    "@emotion/styled": "^11",
```
cindy1u0, line 14: is there a reason why we are using `@emotion/styled` instead of `styled-components`? Or do they come with installing chakra-ui? Cuz if that's the case, we probably should use `@emotion/styled` since they are coupled and prob does the same thing

BrandonLim8890: Upon reading docs charkra-ui asks us to use `@emotion/styled`, and there is no documentation of whether it supports `styled-components`. Someone on github said that chakra-ui won't be supporting `styled-components`, but idk how reliable that is

packages/frontend-v2/utils/theme.ts
aryanshah701: good question! I actually only found 1 design system in the figma([this one](https://www.figma.com/file/CPqFE7T1aJe46zR99n0SPr/GraduateNU-Design-System?node-id=2%3A5)) so the colors are from there, it includes both the blue(from the new designs) and red color(from the current designs). @justineluo and I discussed nailing down a proper new design system in the summer and update the themes to reflect that

packages/frontend-v2/package.json
aryanshah701: @emotion/styled is a peer dependency for chakra, in 99% of cases we should just be able to style things the Chakra way - which is use Chakra components and style through props. This should make it so that everything is consistent with our themes

aryanshah701: that being said if for some reason we do want some custom styling, we'll use @emotion/styled since it's already installed(as a peer dependency), we can update the notion once this is merged 👀

bugsalexander:
[Styling and Theming](https://trello.com/c/tNSb1yop/434-styling-and-theming)

packages/frontend-v2/utils/theme.ts
cindy1u0: ahh okay. I think that could be the old design system but I could be wrong. I guess it's okay we can always change those later

packages/frontend-v2/package.json
cindy1u0: oki sounds good! I guess we will just go with `chakra` + `@emotion/styled` combo!

cindy1u0 (approved):
YAYY U GO GUYS

cindy1u0 (commented):
actually second thought

packages/frontend-v2/package.json
bugsalexander: why does chakra require @emotion/styled and @emotion/react? since this will introduce a lot of different ways to style our components, what's the one we should be using for our codebase?

packages/frontend-v2/package.json
```diff
@@ -9,7 +9,11 @@
...
     "@graduate/common": "workspace:*",
+    "framer-motion": "^6",
```
bugsalexander, line 16: why do we need framer?

packages/frontend-v2/pages/index.tsx
bugsalexander: it sounds like we will be defaulting to chakra ui props over emotion/styled then? we should also be careful to not over-use these, (ie multiple nesting of row/column component props and adding lots of nodes over styling)

bugsalexander (changes requested):
ik u said you would do it after this is merged, but could you do it before? a lot of the platform doc for styling needs to be changed now, so i think it's important to add that as part of the plan for these changes.

additionally, if the main way to style components now will be using the css-in-js props that chakra gives us, it should be a deliberately-made choice. ie, recognize that we will be using a framework with less ubiquity–devs with experience using vanilla CSS or SASS will have to learn something new

bugsalexander:
another thing to think about is not all of our components will be chakra components–how do we apply styles to those? do we use emotion? if so, then now there will be multiple methods of writing styles in our codebase (styled-components-like way using emotion, and via chakra UI props) that devs need to be familiar with before being able to understand code.

packages/frontend-v2/package.json
```diff
@@ -9,7 +9,11 @@
     "lint": "next lint"
   },
   "dependencies": {
+    "@chakra-ui/react": "^1.8.8",
```
bugsalexander, line 12: another thing to think about is not all of our components will be chakra components–how do we apply styles to the non-chakra components? do we use emotion?

if so, then now there will be multiple methods of writing styles in our codebase (styled-components-like way using emotion, and via chakra UI props) that devs need to be familiar with before being able to understand code.

packages/frontend-v2/package.json
```diff
@@ -9,7 +9,11 @@
...
+    "@emotion/react": "^11",
```
bugsalexander, line 13: and then also, im guessing we should use chakra props over emotion to style chakra components? so making that clear in the docs as well. 

when introducing multiple ways to accomplish the same task into a codebase, there need to be much more stringent guidelines for how to go about accomplishing things, so new devs know what the ground rules are. in general, it adds things to learn/read/be familiar with before being able to comfortable contribute to a codebase

packages/frontend-v2/package.json
aryanshah701: honestly no idea but they are simple peer dependencies for chakra. the idea is to only use Chakra and not anything else

packages/frontend-v2/package.json
aryanshah701: peer dependency again, perhaps used by chakra modals and things.

Link to the installation i followed: https://chakra-ui.com/guides/getting-started/nextjs-guide

packages/frontend-v2/package.json
aryanshah701: hmmm... so having used chakra quite a bit, I actually have never needed to write a custom component? I could totaaalllyyy be missing cases where we would need to so let me know!

The whole point of Chakra is to have a single source of truth for styling, even the most basic components are available and can be styled using chakra(divs, paragraphs, etc) almost as extensively as one could style with css. 

The single source of truth makes it much easier to have a unified theme and responsive behaviour that we define in one file and all components behave accordingly. For instance, instead of using a HTML <button> we use <Button> from chakra everywhere, that way if our need to change how a button behaves responsively/sizing(update what "small" means etc), we can do so in the theme file and have it update everywhere. This is different from have a styled.button and using theme values since with Chakra everything is implicitly themed.

Essentially, using Chakra components for even the simplest styles by nature gives us control to every single component on our pages from a single file without having to explicitly use theme values.

packages/frontend-v2/package.json
aryanshah701: definitely!! my point with Chakra was actually to address using multiple ways of styling things in the first place. If you look at the brief Chakra section in the document, it's like the first thing I mention :) 

But yeap I agree that documenting it and making it clear is most important!

packages/frontend-v2/pages/index.tsx
aryanshah701: yeap!

packages/frontend-v2/package.json
bugsalexander: cool, sounds good. we could also take advantage of framer later on for animations, potentially

packages/frontend-v2/package.json
bugsalexander: yeah, makes sense. typically when implementing designs from figma though, small things end up being slightly off (ie, the spacing in design doesn't _quite_ match the spacing defined in theme–though this example is easy to take care of using chakra).

though, it is good to hear that you haven't had to dip into other ways of styling when using chakra

packages/frontend-v2/package.json
bugsalexander: awesome, thank you!

cindy1u0 (approved):
thanks for looking into this again

bugsalexander (approved):
i'd prefer the doc to be done before merging, but if this is blocking other stuff then def good to merge now

BrandonLim8890 (approved):

PR #416, opened 2022-05-07
Compiler-Styled Scraper Rewrite Stage 0.5: URL Filter & create pipeline

bugsalexander:
# Description

Add a small pass to figure out the type for each catalog entry and filter on them.

Also adds a `main.ts` with the full (wip) pipeline for scraping.

[Trello](https://trello.com/c/XKT7qLWO/450-scrapers-adding-catalog-metadata-filtering-out-non-majors)

## Type of change

# How Has This Been Tested?

- [x] Wrote tests

To run tests: run `yarn test filters`. we wrote tests for a handful of majors, minors, concentrations, and a single weird case.

# Checklist:


bugsalexander:
[Scrapers: adding catalog metadata (filtering out non-majors)](https://trello.com/c/XKT7qLWO/450-scrapers-adding-catalog-metadata-filtering-out-non-majors)

packages/scrapers-v2/src/main.ts
```diff
@@ -0,0 +1,21 @@
+import { scrapeMajorLinks } from "./urls/urls";
+import { classifyCatalogEntries } from "./classify/classify";
+import { CatalogEntryType } from "./classify/types";
+import { filterByEntryType, flattenCatalogHierarchy } from "./utils";
+import { fetchAndTokenizeHTML } from "./tokenize/tokenize";
+
+export const runScrape = async () => {
+  try {
+    const hierarchy = await scrapeMajorLinks(2021, 2022);
+    const flattenedUrls = flattenCatalogHierarchy(hierarchy);
```
BrandonLim8890, line 10: Is there a reason why we build up the hierarchy object in `scrapeMajorLinks` and then flatten again? Why not just take the last "parts" from the `scrapeMajorLinks` before building an hierarchy?

packages/scrapers-v2/src/classify/classify.ts
```diff
@@ -0,0 +1,47 @@
+import { ensureLengthAtLeast, loadHTML, parseText } from "../utils";
+import { CatalogEntryType, TypedCatalogEntry } from "./types";
+
+export const classifyCatalogEntries = async (
+  flattenedList: string[],
+): Promise<TypedCatalogEntry[]> => {
+  return await Promise.all(flattenedList.map(addTypeToUrl));
```
BrandonLim8890, line 7: I could be wrong by why await a promise when the return type is a promise. Is it because we are waiting for all async functions to finish within the `Promise.all()`?

packages/scrapers-v2/test/urls.test.ts
```diff
@@ -1,10 +1,10 @@
 // https://catalog.northeastern.edu/undergraduate
 
-import { scrapeMajorLinks } from "../src/urls/scrape_urls";
+import { scrapeMajorLinks } from "../src/urls/urls";
 
 describe("url scraper", () => {
   test("scrape", async () => {
-    jest.setTimeout(15000)
+    jest.setTimeout(15000);
```
BrandonLim8890, line 7: I'm a little confused as to why we need to use a settimeout here. Is it because fetching the major links is async and if it wasn't here, the test function would complete before we get our result?

BrandonLim8890 (commented):
V clean v sexy. I just have a couple questions that I left throughout the review. Similarly, really easy to understand while having the catalog up next to the code, but would be even easier and better with documentation and examples showing why a tab length of 2 == minor/concentration vs 3 == major. But overall was really easy to understand and looks really good!

packages/scrapers-v2/src/main.ts
bugsalexander: it's because in the future, it might be helpful to actually keep the hierarchy information. right now we don't need it so we just flatten it

packages/scrapers-v2/src/classify/classify.ts
bugsalexander: if the promise ends up rejecting, then the stacktrace will originate from this function, because the promise was `await`ed here. if we don't include await, and the promise rejects, the stacktrace will originate from wherever the promise happens to be awaited, making the code harder to debug

packages/scrapers-v2/test/urls.test.ts
bugsalexander: async jest tests have a default limit of 5sec, after which, they auto-fail. this is here because the scraper tests take around ~5s to run (we add some buffer for slow internet connection, etc).

bugsalexander: https://jestjs.io/docs/jest-object#jestsettimeouttimeout <- mentions the default 5s timeout

bugsalexander:
@BrandonLim8890 we will add a readme in the scrapers-v2 directory that includes the required node version, and will also add more docs for this stage specifically (w examples)(

packages/scrapers-v2/test/urls.test.ts
bugsalexander: LOL from looking at the docs we are using this wrong, we should be using this instead: https://jestjs.io/docs/api#testname-fn-timeout -> will fix

jia614 (approved):
i don't have anything additional to add! again it's so clean and easy to understand ty guys!
i think adding more examples to the docs will address any potential issues with clarity like brandon said. otherwise, tests look thorough & work on my end, so lgtm!

packages/scrapers-v2/src/classify/classify.ts
aryanshah701: not awaiting my promises before returning always gets me :(

cindy1u0: tbh im still in the guess and check state even after swe ;))

packages/scrapers-v2/src/classify/classify.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  if (!tabsContainer || tabsContainer.length === 0) {
+    return CatalogEntryType.Unknown;
+  }
+
+  if (tabsContainer.length > 1) {
+    throw new Error(
+      `Expected 1 tab container, but found ${tabsContainer.length}.`
+    );
+  }
```
aryanshah701, line 27: just a thought, it would be nice if we followed either the pattern of throwing errors first and then handling our cases or handling our cases first and then throwing errors consistently. 

For instance, if we were doing errors before and then normal cases then we could have the check of `if (tabs > 3) throw` here instead of at the end. And then within the `tabs.len === 3` case we can do `if middleTabText !== "Program requirements" throw` first and then return instead of the other way around. 

Again, totally ok with handling normal cases first and then throwing errors, would just like to see 1 consistent pattern so that it's easy to follow through the different cases

aryanshah701 (approved):
LGTM!! approving, just have 1 comment before merging tho

packages/scrapers-v2/src/classify/classify.ts
aryanshah701: lmfaoo

BrandonLim8890 (approved):

packages/scrapers-v2/src/classify/classify.ts
bugsalexander: yep, refactored

packages/scrapers-v2/test/urls.test.ts
```diff
@@ -4,8 +4,7 @@ import { scrapeMajorLinks } from "../src/urls/urls";
 
 describe("url scraper", () => {
   test("scrape", async () => {
-    jest.setTimeout(15000);
     const result = await scrapeMajorLinks(2021, 2022);
     expect(result).toMatchSnapshot();
-  });
+  }, 15000);
```
cindy1u0, line 9: edited this part because I think we weren't using setTimeout right before

packages/scrapers-v2/src/urls/urls.ts
```diff
@@ -8,11 +8,10 @@ import { CatalogHierarchy, College, CatalogPath } from "./types";
  * @param end   ending year
  * @returns     a hierarchy of catalog entry links
  */
-export const scrapeMajorLinks = async (start: number, end: number) => {
-  // technically we could specify year via 1 number, but this is ambiguous
-  // ex would "2022" refer to catalog year 2021-2022, or 2022-2023?
-  // instead, we are explicit and require two sequential numbers, like 2021-2022
```
bugsalexander, line 14: wrong merge, add this back

packages/scrapers-v2/src/classify/README.md
```diff
@@ -0,0 +1,36 @@
+# Classify
+
+This folder contains code for the catalog entry classification stage of the scraper pipeline. In this stage, we classify
+catalog entries and annotate them with their type.
+
+## Classification Strategy
+
+We classify catalog entries by looking at the set of tabs (if it exists) at the top of the page. Here is a breakdown of
+the identifiable traits of our currently recognized entry types, by type:
```
cindy1u0, line 9: idk if im tripping but I think there's an extra line break after "breakdown of?" And there are some other ones in the later paragraphs

cindy1u0 (approved):

BrandonLim8890 (approved):
LGTM

PR #419, opened 2022-06-01
Fix CI Build

cindy1u0:
# Description

After https://github.com/sandboxnu/graduatenu/pull/413 was merged, the CI build failed because scrapers-v2's tsconfig is missing some options. Since we are using decorators in `dto-types`, we need to add `experimentalDecorators` to the config (more info see [here](https://www.typescriptlang.org/docs/handbook/decorators.html)). Additionally, we also initialize without having constructors in `dto-types`, so we also need an option for `strictPropertyInitialization` (see [here](https://www.typescriptlang.org/tsconfig#strictPropertyInitialization)). 

Though I am not sure why CI build wasn't catching it before in the branch... unless the dto stuff was merged after. 

## Type of change

# How Has This Been Tested?

CI passing

# Checklist:

### DIFF
--- a/packages/scrapers-v2/tsconfig.json
+++ b/packages/scrapers-v2/tsconfig.json
@@ -16,7 +16,9 @@
     "lib": ["dom", "dom.iterable", "esnext"],
     "isolatedModules": true,
     "noEmit": true,
-    "jsx": "react"
+    "jsx": "react",
+    "strictPropertyInitialization": false,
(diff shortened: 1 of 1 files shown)


BrandonLim8890 (approved):
LGTM

bugsalexander (approved):
at some point, we should really make a common tsconfig and then extend it in each of the pacakages

PR #420, opened 2022-06-03
SWR Custom Hooks for useStudent and usePlan

BrandonLim8890:
# Description

This PR installs `swr`, and adds SWR custom hooks for student and plan. It streamlines the process of getting the `student` and `plan` from the backend by placing it in an SWR hook. This also allows for mutation of these objects, and optimistic updates.

## Type of change

# How Has This Been Tested?

I have created 3 additional pages on the frontend:
1. To act as a [login](http://localhost:3000/login) page.
2. To act as a page that uses [student](http://localhost:3000/testusestudent).
3. To act as a page that uses [plan](http://localhost:3000/useplan).

To test:
- Run `yarn packages/api-client build` to ensure that that the api-client is looking at localhost:3001
- Ensure docker is running
- Run `yarn dev:db:up`
- Run `yarn dev:v2`
- - Delete all instances of "Aryan Shah" on the db using the "Clear DB" button.
- "Register" on the frontend, and create a new plan.
- Test out the various usestudent/useplan links above to see that you are able to access `student` and `plan` when logged in.
- Return back to the main page, logout and test the above links again.
- Ensure that you are redirected to /login as you are not logged in to make those api requests.

# Checklist:


BrandonLim8890:
[SWR Custom hooks](https://trello.com/c/cbKho6A5/427-swr-custom-hooks)

packages/api-client/index.ts
```diff
@@ -32,7 +32,7 @@ class APIClient {
     return responseClass ? plainToInstance(responseClass, res) : res;
   }
 
-  constructor(baseURL = "/api") {
+  constructor(baseURL = "http://localhost:3001/api") {
```
BrandonLim8890, line 35: This would be changed back for prod as both the backend and frontend would be running on the same URL, but for testing purposes the backend and frontend are on different ports.

packages/api-v2/src/main.ts
```diff
@@ -9,6 +9,9 @@ async function bootstrap() {
     logger: ["log", "error", "warn", "debug", "verbose"],
   });
 
+  // Enabled for testing :(
+  app.enableCors();
```
BrandonLim8890, line 13: Enabled CORS for testing, but as I indicated earlier, this won't be needed for prod as both front and back end will both be on the same URL. Thus no requirement for CORS.

rael346 (approved):
LGTM

packages/api-client/index.ts
BrandonLim8890: No longer needed as you can now use http://localhost:3002

packages/frontend-v2/hooks/useLocalStorage.ts
```diff
@@ -0,0 +1,35 @@
+import { useState, Dispatch, SetStateAction } from "react";
+
+const isWindow = typeof window !== "undefined";
```
jia614, line 3: why do we need this check in lines 11/25? might be useful to add a comment here explaining why

packages/frontend-v2/pages/login.tsx
```diff
@@ -0,0 +1,11 @@
+import { NextPage } from "next";
+
+const Login: NextPage = () => {
+  return (
+    <>
+      <h1> You need to login</h1>
```
jia614, line 6: this is probably unnecessary kekw and would probably be updated in a future ticket so feel free to ignore but maybe make this (since its login) chakra heading for the sake of consistency?

jia614 (approved):
its soooOooOoOo clean 🤩 j had two tiny nits and it lgtm yayay

packages/frontend-v2/pages/login.tsx
BrandonLim8890: Although a small change, I don't think it's required since this page is more of a proof of concept to show that the redirect when the user is invalid works. This would all change when we implement the actual login page.

jia614: yaya makes sense

packages/frontend-v2/hooks/utils.ts
```diff
@@ -0,0 +1,15 @@
+import {AxiosError} from "axios";
+import {useRouter} from "next/router";
+
+export const useRedirectUnAuth = () => {
+  const router = useRouter();
+  return (error: AxiosError) => {
+    if (
+      error.response &&
+      (error.response.status === 401 || error.response.status === 403)
```
bugsalexander, line 9: why redirect on 401 or 403? i think it would make more sense to instead check if user is authorized on page load, and if not redirect to login page (something like [this](https://github.com/HackBeanpot/application-portal/blob/9f2ec318cca451d828b8ca651a7e8980f5b5ed19/hooks/useSessionOrRedirect.ts))

packages/frontend-v2/pages/index.tsx
```diff
@@ -10,8 +14,9 @@ const Bomb: React.FC = () => {
 const Home: NextPage = () => {
   const [isClientSideError, setIsClientSideError] = useState(false);
 
-  const [token, setToken] = useState("");
   const [planId, setPlanId] = useState<number>();
+  const [tokenInStorage, setTokenInStorage] = useLocalStorage("token", "");
```
bugsalexander, line 18: it is actually very insecure to store token in localStorage (because it is accessible via JS). i know this is temp code, but just figured id point it out

packages/frontend-v2/pages/index.tsx
```diff
@@ -33,16 +38,29 @@ const Home: NextPage = () => {
     );
   };
 
+  const clearDataBase = async () => {
+    const res = await axios.get("http://localhost:3002/api/students");
+    const uuids = res.data.map((student: GetStudentResponse) => student.uuid);
+    for (const uuid of uuids) {
+      await axios.delete(`http://localhost:3002/api/students/${uuid}`);
+    }
```
bugsalexander, line 46: this is slow, because it doesn't make parallel requests. instead do something like 

```typescript
await Promise.all(uuids.map(uuid => axios.delete(...)));
```

packages/frontend-v2/pages/index.tsx
```diff
@@ -290,137 +316,138 @@ const Home: NextPage = () => {
...
       <div>
         <h2>Logging</h2>
-        <button onClick={() => logger.info("Info log")}>info</button>
-        <button onClick={() => logger.debug("Debug log")}>debug</button>
-        <button onClick={() => logger.warn("Warning log")}>warning</button>
-        <button onClick={() => logger.error("Error log")}>error</button>
+        <Button onClick={() => logger.info("Info log")}>info</Button>
+        <Button onClick={() => logger.debug("Debug log")}>debug</Button>
+        <Button onClick={() => logger.warn("Warning log")}>warning</Button>
+        <Button onClick={() => logger.error("Error log")}>error</Button>
```
bugsalexander, line 450: in the future, no need to make button changes, since this is temp code

packages/frontend-v2/hooks/usePlan.ts
```diff
@@ -0,0 +1,35 @@
...
+  if (axios.isAxiosError(error)) {
+    redirect(error);
+  }
+
+  return {
+    plan,
+    error,
+    mutatePlan,
+    isLoading: !plan && !error,
+  };
```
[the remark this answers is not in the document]
bugsalexander: a trick i like doing is something like this, though idk how the types would work out
```typescript
const { data, mutate, ...rest } = useSwr(...);
return { ...rest, plan: data, mutatePlan: mutate };
```

packages/frontend-v2/hooks/useLocalStorage.ts
```diff
@@ -0,0 +1,37 @@
+import { useState, Dispatch, SetStateAction } from "react";
+
+// NextJS executes server side code first, and then client side. Window object is only present client side,
+// so window could be undefined. This ensures the window object exists before we use it.
+const isWindow = typeof window !== "undefined";
+
+export function useLocalStorage<T>(
```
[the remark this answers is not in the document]
aryanshah701: +1

packages/frontend-v2/hooks/useLocalStorage.ts
```diff
@@ -0,0 +1,37 @@
...
+  key: string,
```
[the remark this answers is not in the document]
aryanshah701: +1

packages/frontend-v2/hooks/useLocalStorage.ts
```diff
@@ -0,0 +1,37 @@
...
+      const valueToStore =
+        value instanceof Function ? value(storedValue) : value;
+      setStoredValue(valueToStore);
+
+      isWindow &&
+      window.localStorage.setItem(key, JSON.stringify(valueToStore));
+    } catch (error) {
+      console.error(error);
+    }
+  };
```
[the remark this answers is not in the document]
aryanshah701: haha yea that's a fair point

packages/frontend-v2/hooks/usePlan.ts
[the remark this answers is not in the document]
aryanshah701: I agree with this, for the type you can check what the return type for useSWR is. That may be easier than having the interface since UsePlanReturn can just be the same type returned by `useSWR<return type of fetcher, error type thrown by fetcher>`. Not quite sure of the generics, you'll have to take a deeper look into it.

packages/frontend-v2/hooks/utils.ts
aryanshah701: ah yea 403 should not redirect to sign-in since that should mean that the user is authenticated just not authorized.

Also, I had suggested this but i think delegating the work of checking whether the user is authorized or not to the component before accessing protected api routes makes sense.

We can remove checks for 401s when accessing protected data in hooks and instead return just return the error. Components can be responsible for ensuring the user is authenticated. If the user is authenticated and we get a 403 then we components can display an appropriate message.

packages/frontend-v2/pages/index.tsx
aryanshah701: lmao yes! it's accessible by any js that runs from our domain making us vulnerable to xss, cookies fix this but gotta make sure same site attribute is set to lax/strict to avoid csrf - something to keep in mind when doing the cookie impl.

packages/frontend-v2/pages/index.tsx
aryanshah701: so the html buttons with chakra don't even look like buttons for some odd reason. I asked brandon to simply change the buttons to the chakra buttons so that the testing ux wasn't completely unusable :/

aryanshah701 (commented):
looks awesome!! good to ship once @bugsalexander comments are addressed :))

packages/frontend-v2/hooks/useLocalStorage.ts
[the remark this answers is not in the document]
BrandonLim8890: Decided not to do this because the hook already ensures that the key is always going to be correct. And since the plans key has an `id` field, it would have made things more complicated. Might end up doing this at a later time tho, but Aryan and I decided not to do this for this PR.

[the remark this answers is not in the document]
bugsalexander: imo storing any form of fetchable data in localstorage is not needed – it's more for the use case of tracking form progress (in onboarding, for example). and what do you mean by "the hook ensures that the key is always going to be correct?"?

packages/frontend-v2/hooks/useLocalStorage.ts
```diff
@@ -1,33 +1,38 @@
...
+  defaultValue: T
+): [T, Dispatch<T>] {
+  const [storedValue, setStoredValue] = useState<T>(defaultValue);
+
+  useEffect(() => {
     try {
-      const item = isWindow && window.localStorage.getItem(key);
-      return item ? JSON.parse(item) : initialValue;
+      const item = window.localStorage.getItem(key);
+      setStoredValue(item ? JSON.parse(item) : defaultValue);
```
bugsalexander, line 24:
```suggestion
      item ?? setStoredValue(JSON.parse(item));
```

no need to re-set value as defaultValue, since it will already be default value

packages/frontend-v2/hooks/useLocalStorage.ts
```diff
@@ -1,33 +1,38 @@
...
+// These are here for testing purposes.
+export enum LocalStorageKey {
+  token = "token"
+}
 
+/**
+ * Using this hook allows for any component to access localstorage.
+ * Accessing the hook allows for the functionality of setting and getting a value in localstorage.
+ * @param key Key for the localstorage value.
+ * @param defaultValue Default value returned if the hook cannot access the value in localstorage.
```
bugsalexander, line 13: pretend that someone reading the comment will only read the comment to figure out what this hook does, and not look at the code (because this is often the case).

~~in this case, "if the hook cannot access the value" is not clear enough for me–you should specify that (1) it~~

sorry i had a better idea midsentence. i think having this hook _not_ take a default value, but then return `null | T` is WAY better design.

- (1) this will FORCE callee to handle the null case, which is good because this is a pretty common use case, and then they can choose exactly how to handle it with a nullcheck, and
- (2) they dont have to understand the weirdness of us only _defaulting_ to the default value, but not actually setting it, etc.

thoughts?

packages/frontend-v2/hooks/useLocalStorage.ts
```diff
@@ -1,33 +1,38 @@
...
-      const item = isWindow && window.localStorage.getItem(key);
-      return item ? JSON.parse(item) : initialValue;
+      const item = window.localStorage.getItem(key);
+      setStoredValue(item ? JSON.parse(item) : defaultValue);
     } catch (error) {
       console.error(error);
-      return initialValue;
     }
-  });
+  }, [])
```
bugsalexander, line 28: defaultValue missing here as a dependency, but u can ignore it if we decide to go with the design i proposed above

packages/frontend-v2/hooks/useLocalStorage.ts
```diff
@@ -1,33 +1,38 @@
...
 
-  const setValue = (value: T | ((value: T) => T)) => {
+  const setValue = (value: T) => {
```
bugsalexander, line 30: so according to the rules of hooks, refs should be stable, meaning that re-renders should not cause things to change that do not need to change. however whenever this hook rerenders, this function will be re-evaluated, producing a new ref, and whatever things have this setter in their dependency array will re-calculate needlessly. so in theory, this should be memoized with `useCallback` but this is probably overkill, so feel free to ignore this.

relevant docs reference: https://reactjs.org/docs/hooks-reference.html#usecallback

the use case i'm referring to here is where it says:

> This is useful when passing callbacks to optimized child components that rely on reference equality to prevent unnecessary renders (e.g. shouldComponentUpdate).

packages/frontend-v2/hooks/useLocalStorage.ts
```diff
@@ -1,33 +1,38 @@
...
-        value instanceof Function ? value(storedValue) : value;
-      setStoredValue(valueToStore);
-
-      isWindow &&
-      window.localStorage.setItem(key, JSON.stringify(valueToStore));
+      setStoredValue(value);
+      window?.localStorage.setItem(key, JSON.stringify(value));
     } catch (error) {
-      console.error(error);
+      console.log(error);
```
bugsalexander, line 35: why not console.error?

packages/frontend-v2/hooks/usePlan.ts
```diff
@@ -1,35 +1,34 @@
-import useSWR, { SWRResponse } from "swr";
+import useSWR, {KeyedMutator, SWRResponse} from "swr";
```
bugsalexander, line 1: formatting xDDD

alright scrap telling u later, this is how to enable prettier format on save:

1. go to webstorm `preferences`
2. go to `plugins`. type in `prettier`. install it if it's not already installed (bundled)
3. search `prettier` in the preferences search bar (top left).
4. click on result under `languages and frameworks > javascript > prettier`. 
5. for `prettier package`, click dropdown, and select the version installed in the `graduatenu` folder (should be version 2.6.x roughly). 
6. make sure `run for files` looks something like this: `{**/*,*}.{js,ts,jsx,tsx,md}` (optional for u whether it includes md or not)
7. check the `on reformat code` action
8. check the `on save` action
9. click `apply` and then click `ok` in the bottom right.

for extra stuff on save:
1. from the above dialogue, click on `all actions on save`
2. check `optimize imports` and `run code cleanup`. `run prettier` should already be checked. i dont like including eslint because it somtimes takes more than a second to run (too long for me)
4. click `apply` and `save` again

packages/frontend-v2/pages/testuseplan.tsx
```diff
@@ -1,26 +1,50 @@
...
 
-  const {plan, error, isLoading} = usePlan(planInStorage, tokenInStorage);
+  const {student, error: studentError, isLoading: studentIsLoading} = useStudentWithPlans(tokenInStorage);
+  if (studentError) {
+    return (
+      <h1>student could not be fetched</h1>
+    )
+    // handle error
+  } else if (studentIsLoading) {
+    // handle loading
```
bugsalexander, line 18: return something here! something like <div> loading</div>

bugsalexander (approved):
last api change i promise. after these are resolved u are good to ship

packages/frontend-v2/hooks/useLocalStorage.ts
```diff
@@ -0,0 +1,35 @@
...
+      setStoredValue(valueToStore);
+
+      isWindow &&
+      window.localStorage.setItem(key, JSON.stringify(valueToStore));
+    } catch (error) {
+      console.error(error);
+    }
+  };
+
+
```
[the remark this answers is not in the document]
cindy1u0: I think Brandon did try to add in the remove value function last week during standup, but there were some type mismatches. Later decided not to add it because the type mismatch was kind of annoying to fix and didn't seem worth to spend more time on since we might not need it for now. But could a comment on that

packages/frontend-v2/hooks/useStudentWithPlans.ts
```diff
@@ -0,0 +1,34 @@
...
+  student?: GetStudentResponse;
+}
+
+/**
+ * Returns the student with plan using SWR.
+ * @param jwt JWT for authentication to fetch the data from db.
+ * Will later be removed when we switch to cookies.
+ */
+export function useStudentWithPlans(jwt: string): UseStudentReturn {
+  const key = `api/students/me`;
```
cindy1u0, line 20: idk it's any priority for now since it's just for testing. I think later on it might be better to make a util function to construct the key for hooks. It's just easier to keep track of.

packages/frontend-v2/pages/index.tsx
```diff
@@ -33,16 +37,27 @@ const Home: NextPage = () => {
     );
   };
 
+  const clearDataBase = async () => {
```
cindy1u0, line 40: I assume this will be removed later?

packages/frontend-v2/pages/testuseplan.tsx
```diff
@@ -0,0 +1,50 @@
...
+  }
+  return (
+    <>
+      { student?.plans.map(plan => (
```
cindy1u0, line 22: Just a suggestion: I would even add a case to check if `student == null`, then can display something like "No student data". But up to you if you want to change it.

packages/frontend-v2/pages/testuseplan.tsx
```diff
@@ -0,0 +1,50 @@
...
+    )
+    // handle error
+  } else if (isLoading) {
+    // handle loading
+  }
+  console.log(plan);
+
+  return (
+    <>
+      {plan ? <h1> You could usePlan!</h1> : <h1>Plan could not have been fetched</h1>}
```
cindy1u0, line 44: or can just do 
```
<>
  <h1> you could usePlan!</h1>
  <p>{plan.id}</p>
<>
```
instead of having to use optional chaining here since you already check earlier. Again, up to you if you want to change it :)

cindy1u0 (approved):
Overall LGTM! Just some small comments

packages/frontend-v2/hooks/usePlan.ts
BrandonLim8890: you're a life saver

PR #422, opened 2022-06-07
Landing page

jia614:
landing

# Description
This PR implements the landing page according to Linda's designs found [here](https://www.figma.com/file/yUflVFVxjMynm7gGymLpQY/%F0%9F%8F%97-Current-Work?node-id=37%3A150).

## Type of change
- [X] This requires a run of `yarn install`

# How Has This Been Tested?

- [X] Tested responsiveness by playing around with window size.

# Checklist:


packages/frontend-v2/package.json
```diff
@@ -12,6 +12,7 @@
     "@chakra-ui/react": "^1.8.8",
     "@emotion/react": "^11",
     "@emotion/styled": "^11",
+    "@fontsource/montserrat-alternates": "^4.5.9",
```
jia614, line 15: font for Graduate logo

packages/frontend-v2/pages/index.tsx
```diff
@@ -1,429 +1,86 @@
...
-            Register
-          </button>
+const Banner = (): JSX.Element => {
+  return (
+    <Flex pt={{ desktop: "75px", laptop: "50px", tablet: "40px" }} pb={{ desktop: "175px", laptop: "150px", tablet: "100px"}}>
+      <HStack direction="row" spacing={{ desktop: "100px", laptop: "75px", tablet: "50px" }} justifyContent="center">
+        <Image boxSize={{ desktop: "550px", laptop: "500px", tablet: "400px" }} src="/husky.svg" alt="husky"/>
+        <Flex w="35%" flexDirection="column" alignItems="center">
+          <Box>
+            <Heading fontSize={{ desktop: "7xl", laptop: "6xl", tablet: "5xl"}} color="primary.main">
```
jia614, line 31: the `size` prop range for chakra headings was not quite as large as i needed it to be here, so i applied the breakpoints using the `fontSize` prop instead

packages/frontend-v2/pages/index.tsx
```diff
@@ -1,429 +1,86 @@
...
-        <button onClick={() => logger.info("Info log")}>info</button>
-        <button onClick={() => logger.debug("Debug log")}>debug</button>
-        <button onClick={() => logger.warn("Warning log")}>warning</button>
-        <button onClick={() => logger.error("Error log")}>error</button>
-      </div>
-    </>
+const InfoSection = ({ title, desc }: InfoSectionProps): JSX.Element => {
+  return (
+    <Box w="55%">
+      <Heading pt="10%" fontSize={{ desktop: "3xl", laptop: "2xl", tablet: "xl" }} color="blue.700">
```
jia614, line 76: i found that the auto-responsiveness of the `size` prop didn't scale down too well as i decreased window size, so i opted for `fontSize`/explicit breakpoints instead

packages/frontend-v2/utils/component_styles/buttons.ts
```diff
@@ -0,0 +1,39 @@
+import { ComponentStyleConfig } from "@chakra-ui/react";
+
```
jia614, line 2: i wasn't sure if i should've created custom button components, but it seems like the best practice for chakra component styling looks something like this according to the docs:

https://chakra-ui.com/docs/styled-system/theming/component-style
https://chakra-ui.com/docs/styled-system/theming/customize-theme

i decided to go with this design since i'm not too sure what we want to do regarding component styling reuse. i think this is subject to change depending on what Aryan decides in the chakra design docs, but if anyone else has any thoughts pls lmk : D

packages/frontend-v2/utils/theme.ts
```diff
@@ -76,8 +81,29 @@ const space = {
...
 
+/**
+ * Customized component styles can be configured and passed to the theme provider
+ */
+const components = {
+  Button,
+}
+
+// Allows us to reuse common styles across pages
+const layerStyles = {
```
jia614, line 92: [this](https://chakra-ui.com/docs/styled-system/features/text-and-layer-styles) allows styling combinations to be reused, but i think putting this & other reusable combinations here could cause the theme file to get cluttered. if this is the right way to go, it might be worth it to place these objects in a separate file and import them? or it might be better to make this its own custom component... tbh im not sure lol

packages/frontend-v2/pages/_app.tsx
```diff
@@ -5,6 +5,7 @@ import { theme } from "../utils";
 import "react-toastify/dist/ReactToastify.min.css";
 import { ToastContainer } from "react-toastify";
 import { ErrorBoundary } from "../components";
+import "@fontsource/montserrat-alternates";
```
bugsalexander, line 8: for anyone else, looks like this is the preferred way to install fonts for chakra https://chakra-ui.com/docs/styled-system/recipes/using-fonts probably good to add this to docs @aryanshah701

packages/frontend-v2/components/index.ts
```diff
@@ -1 +1 @@
-export * from "./Error";
+export * from "./Error";
```
bugsalexander, line 1: you can install the prettier extension (by "prettieR", has 21m downloads) and then enable "format on save" in preferences, and this should get fixed automatically

bugsalexander (approved):
zero clue how chakra works but LGTM

packages/frontend-v2/pages/_app.tsx
aryanshah701: yeap, will do!

packages/frontend-v2/utils/theme.ts
aryanshah701: I love the thought process here! I do feel that the flex header container is better suited as a react component though since header container styles are specific to headers for pages. A single page at most has 1 header and a header is always a Flex component.

Layered styles are nice when we have a group of styles that we would like to use for across multiple components. For example, a set of styles for "selected" that could be applied to Boxes, Buttons etc. would be an awesome layered style.

For here though, we can have a header container component. It can go in the components directory since it seems reusable :)

packages/frontend-v2/utils/component_styles/buttons.ts
aryanshah701: This is awesome! For purely stylistic components like Buttons we def can use this configuration like customization that allows us to create our own UI library/design system.

packages/frontend-v2/utils/theme.ts
```diff
@@ -1,4 +1,5 @@
 import { extendTheme } from "@chakra-ui/react";
+import { Button } from "./component_styles/buttons"
```
aryanshah701, line 2: You were right, the theme file is actually getting quite cluttered now. We can break it up and make theme its own directory under utils. Checkout out the [scaling out design section in the docs](https://chakra-ui.com/docs/styled-system/theming/customize-theme#scaling-out-your-project)!

According to the above example, we can move the components/Button into the new theme directory, and the current theme file will be theme/index.ts.

packages/frontend-v2/components/Header/header.tsx
```diff
@@ -0,0 +1,11 @@
...
+
+export const Logo = (): JSX.Element => {
+  return (
+    <HStack flexDirection="row" alignItems="center" spacing="2px">
+      <Heading size="lg" color="primary.main" fontFamily="logo">Graduate</Heading>
+      <Heading size="xl" color="blue.700" pb="2%">NU</Heading>
+    </HStack>
+  );
+
+}
```
aryanshah701, line 11: Let's rename this to Logo.tsx(capital L) for convention. We're following the convention of naming files based on the primary export of the file.

aryanshah701: This directory is also a good place to put the HeaderContainer.tsx file. 

Be sure to have a barrel file(index.ts) to export everything in the Header directory and then export everything from the header directory from components/index.ts. 

This allows our imports to be much cleaner - "../components", instead of "../components/Header/HeaderContainer". It also allows us to group all of our imports from the components directory so we can do things like "import {Header, Logo, Button} from "../components". Finally it allows us to reorganize the directory structure within components without having any impact outside components :)

packages/frontend-v2/pages/index.tsx
```diff
@@ -1,455 +1,85 @@
 import type { NextPage } from "next";
-import { useState } from "react";
```
aryanshah701, line 2: VERY CLEAN!

packages/frontend-v2/pages/index.tsx
aryanshah701: good thinking!

packages/frontend-v2/pages/index.tsx
```diff
@@ -1,455 +1,85 @@
...
+          pb={{ desktop: "125px", laptop: "130px", tablet: "105px" }} backgroundColor="blue.50">
+      <VStack direction="column">
+        <Heading mb={{ desktop: "95px", laptop: "75px", tablet: "65px" }} size="2xl" color="blue.700">How It Works</Heading>
+        <SimpleGrid columns={3} justifyItems="center" pl="5%" pr="5%"> 
+          <Image boxSize={{ desktop: "250px", laptop: "200px", tablet: "150px" }} pt="5%" src="/landing_start.svg" alt="Start" />
+          <Image boxSize={{desktop: "250px", laptop: "200px", tablet: "150px" }} pl="5%" src="/landing_personalize.svg" alt="Personalize" />
+          <Image boxSize={{ desktop: "250px", laptop: "200px", tablet: "150px" }} pt="5%" src="/landing_graduate.svg" alt="Graduate" />
+          <InfoSection title="Start" desc="Just answer a couple questions and get started with a multi-year plan for your classes."/>
+          <InfoSection title="Personalize" desc="Pick the classes you want. We'll take care of NU Path, pre-requisites, and everything in between."/>
+          <InfoSection title="Graduate" desc="Build a plan of study that lets you graduate faster, with better classes, and a lot less headaches."/>
```
aryanshah701, line 60: Super job making this a component. We can go a small step further and remove the data from the return state and put it into its own object somewhere in the component. It declutters the static data from the return statements and also centralized it so that we don't have to look for it in the return :)

```ts
const infoSectionData = [{
    title: "...",
    description: "..."
},
{
    tile: "...",
    description: "..."
},
...
]


...

return (
    ...
   {infoSectionData.map(info => <InfoSection title={info.title} description={info.description} />}
)
```

packages/frontend-v2/pages/index.tsx
```diff
@@ -1,455 +1,85 @@
...
+const Info = (): JSX.Element => {
+  return (
+    <Box pt={{ desktop: "95px", laptop: "100px", tablet: "75px" }} 
+          pb={{ desktop: "125px", laptop: "130px", tablet: "105px" }} backgroundColor="blue.50">
+      <VStack direction="column">
+        <Heading mb={{ desktop: "95px", laptop: "75px", tablet: "65px" }} size="2xl" color="blue.700">How It Works</Heading>
+        <SimpleGrid columns={3} justifyItems="center" pl="5%" pr="5%"> 
+          <Image boxSize={{ desktop: "250px", laptop: "200px", tablet: "150px" }} pt="5%" src="/landing_start.svg" alt="Start" />
+          <Image boxSize={{desktop: "250px", laptop: "200px", tablet: "150px" }} pl="5%" src="/landing_personalize.svg" alt="Personalize" />
+          <Image boxSize={{ desktop: "250px", laptop: "200px", tablet: "150px" }} pt="5%" src="/landing_graduate.svg" alt="Graduate" />
```
aryanshah701, line 57: we can make this a component cause the responsive styles and padding is repeated quite a bit :)

packages/frontend-v2/pages/index.tsx
```diff
@@ -1,455 +1,85 @@
...
-    <>
-      <h1>GraduateNU Landing Page:</h1>
-      <div>
-        <h2>Testing api calls on backend v2</h2>
-        <p>All responses will be logged to the console</p>
-
-        <Button onClick={clearDataBase}>Clear DB</Button>
+    <Flex layerStyle="headerContainer">
+      <Logo />
+      <Button size="sm"> Sign In</Button>
```
aryanshah701, line 19: Is the space before "Sign In" supposed to be there?

aryanshah701 (changes requested):
I loved this PR! You really went above and beyond with best practices for chakra, responsiveness and clean code. Learning a new technology, having no guidelines and figuring out solid ways of achieving things is very impressive :)

Have a few comments, ping me when they're address and then it's good to merge!

packages/frontend-v2/components/Header/header.tsx
aryanshah701: Note: If there are any imports from the components directory, be sure to import directly from "../components" instead of the specific component file to reap the benefits of barrels. 

Also we should def not overuse barrels cause it's added boilerplate, but i think they're useful for the components and utils directory since those can get quite nested

packages/frontend-v2/pages/index.tsx
rael346: [This](https://chakra-ui.com/docs/components/typography/heading#changing-font-size) might be the reason why

rael346 (approved):
LGTM

BrandonLim8890 (approved):
Overall looks really good! Just address Aryan's comments and we should be good! I only have 1 opinion regarding px vs rem, but others can also chime in and give their takes. 

Basically, I personally think rem is a WAY better tool for padding/margins than px simply because of user accessibility. If a page were designed with rem for text, but px for margins/padding, then when a user zooms in (because text may be small and difficult to read), only the text will scale up in size, not the margins/padding. However, if a site is written with rem for all padding/margins, then the website fluidly scales up and down with zoom, or in accordance to the client's default font size. I hope that kind of makes sense? You can read more about it [here](https://stackoverflow.com/questions/11799236/should-i-use-px-or-rem-value-units-in-my-css#:~:text=Respecting%20a%20user%27s,scale%20up%20fluidly.).

packages/frontend-v2/utils/theme.ts
jia614: ok sounds good!!

packages/frontend-v2/components/Header/header.tsx
jia614: sounds good :)

packages/frontend-v2/pages/index.tsx
```diff
@@ -1,455 +1,85 @@
...
-              });
-              console.log(student);
-              setTokenInStorage(student.accessToken!);
-            }}
-          >
-            Register
-          </Button>
+const Banner = (): JSX.Element => {
+  return (
+    <Box pt={{ desktop: "75px", laptop: "50px", tablet: "40px" }} pb={{ desktop: "175px", laptop: "150px", tablet: "100px"}}>
```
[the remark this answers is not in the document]
jia614: ooooh this is very cool & it makes sense, thank u!

packages/frontend-v2/pages/index.tsx
```diff
@@ -1,454 +1,190 @@
...
-              setTokenInStorage("");
-              console.log("Logged out, token reset");
-            }}
-          >
-            Logout
-          </Button>
+const Info = (): JSX.Element => {
+  const infoImageData = [
+    {
+      imageSource: "/landing_start.svg",
```
cindy1u0, line 87: seems like this `infoImageData` can be combined with `infoSectionData`. In this case, you can use one `map` down there and use react fragments `<></>` to combine the image and the description instead of having two maps.

packages/frontend-v2/pages/index.tsx
```diff
@@ -1,454 +1,190 @@
...
+              description={info.description}
+            />
+          ))}
+        </SimpleGrid>
+      </VStack>
+    </Box>
+  );
+};
+
+interface InfoImageProps {
```
cindy1u0, line 152: I would probably bring all the interfaces before all the elements

cindy1u0 (approved):
Overall lgtm after other people's comments! Just had some minor additional comments

aryanshah701 (approved):
Looks good!

packages/frontend-v2/pages/index.tsx
jia614: using the fragments made the order kinda weird, but i realized i could just make InfoSection (both image/text) its own component and use the single map... ty for the help :')))))

PR #426, opened 2022-06-27
Updates the Backend to Support Onboarding Flow

BrandonLim8890:
# Description

This PR is an update to the current `api-v2` backend where it now follows through with the Onboarding and Signup flow discussions held during component design. In more specific the implemented flow is as follows:

User signs up 
-> Information saved in backend, with all fields apart from `email`, `password`, and `uuid` set to null
-> Client side, user is redirected to onboarding
-> PATCH request to update the user after onboarding by sending a request to `/api/students/me/onboard`
-> Backend route requires for all fields to be present, then will update the user in db accordingly as well as set `isOnboarded` field of the user to be true

This implementation can be further looked at [here](https://www.notion.so/sandboxnu/Signup-Login-Design-dce4ba0d1c9f4eb3920490b9a43e1743).

## Type of change

# How Has This Been Tested?

Run the appropriate calls to the backend to test that:
1. Registration of a user only populates `email` and `password`, other fields are null and `isOnboarded` is set to `false`.
2. Perform a PATCH request to the onboard route, and ensure that the user's fields are now all populated, and `isOnboarded` is set to `true`

The POSTMAN collection for this can be found [here](https://github.com/sandboxnu/graduatenu/files/8988462/GraduateNU-APIv2.json.zip).

# Checklist:

### DIFF
--- a/packages/api-v2/src/student/student.controller.ts
+++ b/packages/api-v2/src/student/student.controller.ts
@@ -73,7 +73,7 @@ export class StudentController {
   }
 
   @UseGuards(JwtAuthGuard)
-  @Patch("onboard")
+  @Patch("me/onboard")
   async onBoard(
     @Req() req: AuthenticatedRequest,
     @Body() updateStudentDto: OnboardStudentDto
--- a/packages/api-v2/src/student/student.controller.ts
+++ b/packages/api-v2/src/student/student.controller.ts
@@ -1,28 +1,29 @@
 import {
-  Controller,
-  Get,
-  Post,
-  Body,
-  Patch,
-  Param,
-  Delete,
-  ParseUUIDPipe,
   BadRequestException,
-  UseGuards,
-  Req,
-  Query,
-  ParseBoolPipe,
+  Body,
+  Controller,
   DefaultValuePipe,
+  Delete,
+  Get,
+  Param,
+  ParseBoolPipe,
+  ParseUUIDPipe,
+  Patch,
+  Post,
+  Query,
+  Req,
+  UseGuards,
 } from "@nestjs/common";
 import { StudentService } from "./student.service";
 import { JwtAuthGuard } from "src/guards/jwt-auth.guard";
 import { DevRouteGuard } from "src/guards/dev-route.guard";
 import { AuthenticatedRequest } from "src/auth/interfaces/authenticated-request";
 import {
-  GetStudentResponse,
-  UpdateStudentResponse,
   CreateStudentDto,
+  GetStudentResponse,
+  OnboardStudentDto,
   UpdateStudentDto,
+  UpdateStudentResponse,
 } from "../../../common";
 
 @Controller("students")
@@ -56,6 +57,32 @@ export class StudentController {
     const updateResult = await this.studentService.update(
       uuid,
       updateStudentDto
+    );
+
+    if (!updateResult) {
+      throw new BadRequestException();
+    }
+
+    const student = await this.studentService.findByUuid(uuid, true);
+
+    if (!student) {
+      throw new BadRequestException();
+    }
+
+    return student;
+  }
+
+  @UseGuards(JwtAuthGuard)
+  @Patch("onboard")
+  async onBoard(
+    @Req() req: AuthenticatedRequest,
+    @Body() updateStudentDto: OnboardStudentDto
+  ): Promise<UpdateStudentResponse> {
+    const uuid = req.user.uuid;
+    const updateResult = await this.studentService.update(
+      uuid,
+
+      { ...updateStudentDto, isOnboarded: true }
     );
 
     if (!updateResult) {
--- a/packages/api-v2/src/student/student.controller.ts
+++ b/packages/api-v2/src/student/student.controller.ts
@@ -79,11 +79,10 @@ export class StudentController {
     @Body() updateStudentDto: OnboardStudentDto
   ): Promise<UpdateStudentResponse> {
     const uuid = req.user.uuid;
-    const updateResult = await this.studentService.update(
-      uuid,
-
-      { ...updateStudentDto, isOnboarded: true }
-    );
+    const updateResult = await this.studentService.update(uuid, {
+      ...updateStudentDto,
+      isOnboarded: true,
+    });
 
     if (!updateResult) {
       throw new BadRequestException();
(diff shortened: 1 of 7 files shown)


packages/common/src/dto-types.ts
```diff
@@ -134,67 +148,46 @@ export class CreateStudentDto {
   concentration?: string;
 }
 
-export class UpdateStudentDto {
-  @IsOptional()
+export class OnboardStudentDto {
```
BrandonLim8890, line 151: A lot of changes here, but basically I make all the fields for a user REQUIRED for OnboardStudentDto as all fields are required if the student has finished onboarding

packages/api-v2/src/student/student.controller.ts
```diff
@@ -71,6 +72,31 @@ export class StudentController {
...
 
+  @UseGuards(JwtAuthGuard)
+  @Patch("me/onboard")
+  async onBoard(
+    @Req() req: AuthenticatedRequest,
+    @Body() updateStudentDto: OnboardStudentDto
+  ): Promise<UpdateStudentResponse> {
+    const uuid = req.user.uuid;
+    const updateResult = await this.studentService.update(uuid, {
+      ...updateStudentDto,
```
bugsalexander, line 83: @aryanshah701 will the validation pipe remove extra fields from the dto? if not, we should explicitly pull the named fields we want from this, to avoid implicitly storing random stuff/overwriting fields in our db.

packages/api-v2/src/student/student.controller.ts
```diff
@@ -71,6 +72,31 @@ export class StudentController {
...
+      isOnboarded: true,
+    });
+
+    if (!updateResult) {
+      throw new BadRequestException();
```
bugsalexander, line 88: internal server error

packages/api-v2/src/student/student.controller.ts
```diff
@@ -71,6 +72,31 @@ export class StudentController {
...
+    }
+
+    const student = await this.studentService.findByUuid(uuid, true);
+
+    if (!student) {
+      throw new BadRequestException();
```
bugsalexander, line 94: internal server error

packages/common/src/dto-types.ts
```diff
@@ -134,67 +148,46 @@ export class CreateStudentDto {
...
-  @IsOptional()
-  @IsNotEmpty()
   @IsString()
-  password?: string;
+  nuid: string;
 
-  @IsOptional()
   @IsInt()
   @Min(1898) // the year NEU was established
-  @Max(3000) // will GraduateNU last beyond year 3000?!
```
bugsalexander, line 160: if this was accidentally removed by a replace all, in the future, don't use replace all – it's dangerous because it can unintentionally change code and cause errors later on

packages/api-v2/src/student/student.controller.ts
```diff
@@ -71,6 +72,31 @@ export class StudentController {
     return student;
   }
 
+  @UseGuards(JwtAuthGuard)
```
bugsalexander, line 75: i don't remember: does this guard make sure that the logged in user matches their uuid?

packages/api-v2/src/student/student.service.ts
```diff
@@ -71,7 +71,7 @@ export class StudentService {
 
   async update(
     uuid: string,
-    updateStudentDto: UpdateStudentDto
+    updateStudentDto: Partial<Student>
```
bugsalexander, line 74: if all the fields in `Student` are optional now, then what's the difference between this and `Partial<Student>`?

packages/common/src/dto-types.ts
```diff
@@ -76,20 +76,34 @@ export class UpdatePlanDto {
 }
 
 export class CreateStudentDto {
```
bugsalexander, line 78: if this is only used in the Signup api, should change to something like SignUpDto (instead of create-student)

packages/common/src/dto-types.ts
```diff
@@ -76,20 +76,34 @@ export class UpdatePlanDto {
...
+  @IsEmail()
+  email: string;
+
   @IsNotEmpty()
   @IsString()
-  fullName: string;
+  password: string;
+}
+
+export class UpdateStudentDto {
```
bugsalexander, line 88: is this used anywhere? i saw that the update endpoint now takes a `Partial<Student>`

packages/common/src/dto-types.ts
bugsalexander: can you use [Required<T>](https://www.typescriptlang.org/docs/handbook/utility-types.html#requiredtype)?

bugsalexander (commented):
looks good, but haven't tested it yet

packages/api-v2/src/student/student.controller.ts
aryanshah701: this guard simply ensures that the JWT is valid and that the uuid stored in the JWT is valid. It then fetches the student corresponding to the uuid stored in the JWT and adds it to the request.

packages/common/src/dto-types.ts
BrandonLim8890: Unless I misunderstood the docs, this isn't possible since these are classes and not types/interfaces. We require classes for class-validator.

packages/api-v2/src/student/student.controller.ts
```diff
@@ -71,6 +73,31 @@ export class StudentController {
...
+  @Patch("me/onboard")
+  async onBoard(
+    @Req() req: AuthenticatedRequest,
+    @Body() onboardStudentDto: OnboardStudentDto
+  ): Promise<UpdateStudentResponse> {
+    const uuid = req.user.uuid;
+    const updateResult = await this.studentService.update(uuid, {
+      ...onboardStudentDto,
+      isOnboarded: true,
```
cindy1u0, line 85: hmm just thinking - does it count as `onboarded` if a student only goes through one step of onboarding? Do we want to say they are only onboarded if all fields have been filled

cindy1u0 (approved):
Tested too overall LGTM :D just have a quick question

packages/api-v2/src/student/student.controller.ts
cindy1u0: oh hm unless we only send this request at the very end of the onboarding then nvm

BrandonLim8890: Yeeeee!! we only send the last request at the very end, which is why in the dto all fields are required.

PR #431, opened 2022-07-13
Store JWT Token in a cookie

cindy1u0:
# Description

This PR changes JWT storing strategy. Instead of storing it in localStorage, we now store the token in a cookie. The following is some other changes that come with the PR:
1. Add logout endpoint to clear cookies
2. Remove localStorage in `index.ts` and other test files that brandon made before
3. Remove `accessToken` field from student because we no longer need to expose that to the client
4. Remove all required `jwt` parameter from endpoint in `api-client`

## Type of change

Might also need to run `build` again in `api-client` to see new changes.

# How Has This Been Tested?

## Test A
Added "logout" and "getMe" buttons locally to test. In the flow, when you sign in, there is a cookie generated in storage. Then we can successfully call `getMe` since it's guarded by JWTAuth. When you log out, the cookie has been removed. Then `getMe` will generate an unauthorized error.

https://user-images.githubusercontent.com/45154128/178666793-b9b1ee38-383d-42d7-bac5-5253cf44a624.mov

### Note
There is a weird warning in firefox when trying to remove cookie. Firefox rejects past cookies, but it is a known bug. Maybe still is...?
[Bug](https://stackoverflow.com/questions/65038173/cannot-remove-a-cookie-firefox-rejecting-cookies-from-the-past)
[Bug Report](https://bugzilla.mozilla.org/show_bug.cgi?id=1676651)

## Test B
https://go.postman.co/workspace/API-V2~7dfc1e4f-7b59-4899-8c38-db7a193ccc11/collection/14551215-d80fa291-e79b-4721-b7d2-b179715ee19a?action=share&creator=19576221

Use the postman collection to test:
1. Login Student -> copy cookie value
5. Use other endpoints and add `Cookie` to headers, and paste the value

# Checklist:

# Moving Forward

## Refresh Token
JWT Storing Strategy:
1. Store your access token in localStorage (and refresh token in either localStorage or httpOnly cookies): the access token is prone to be stolen from an XSS attack.
2. Option 2: Store your access token and refresh token in httpOnly cookie: prone to CSRF but can be mitigated, a bit better in terms of exposure to XSS.
6. Option 3: Store your access token in memory and refresh token in httpOnly cookie: safe from CSRF, a bit better in terms of exposure to XSS.

Currently we are using option 2 but without refresh token. The purpose of having a refresh token is that we can prevent access token from expiring and continuously “refresh” it. We can think about implementing this later on but it may not be necessary now.

To implement refresh token,
1. Add a RefreshStrategy
2. Override validate to validateRefreshToken
3. Have an endpoint to generate refreshToken
7. Add refreshToken along with accessToken to cookie 

Resource: https://wanago.io/2020/09/21/api-nestjs-refresh-tokens-jwt/


packages/api-v2/src/auth/auth.controller.ts
```diff
@@ -11,34 +14,67 @@ import {
...
 } from "../../../common";
+import { Response } from "express";
 
 @Controller("auth")
 export class AuthController {
   constructor(private readonly authService: AuthService) {}
 
   @Post("register")
   public async register(
+    @Res({ passthrough: true }) response: Response,
```
cindy1u0, line 25: we need to add `passthrough` here because since we need to set cookie in response, when we add @Res, we will need to handle the response ourselves. [See here in doc](https://docs.nestjs.com/controllers#appendix-library-specific-approach)

cindy1u0:
https://trello.com/c/KB3vXCNH/444-authorization-header-cookies

cindy1u0:
[Authorization Header -> Cookies](https://trello.com/c/KB3vXCNH/444-authorization-header-cookies)

aryanshah701 (approved):
Awesomeeeee, looks good to me!

PR #437, opened 2022-07-24
Create Signup and Login Workflows (No design element)

BrandonLim8890:
# Description

This PR (LONG OVERDUE) creates a login and signup page for Frontend V2. Both pages make use of class-validator for validation, react-hook-form and the API-Client package. Upon logging in, the user will be redirected to `/home` or `/onboarding` depending on if they have completed the onboarding process.

## Type of change

# How Has This Been Tested?

Ran through workflows myself, and checked that the database is updated via POSTMAN requests to list all students.

# Checklist:


babel.config.js
```diff
@@ -15,7 +15,10 @@ module.exports = function (api) {
           ],
           "@babel/preset-typescript",
         ],
-        plugins: [["@babel/plugin-proposal-decorators", { version: "legacy" }]],
+        plugins: [
```
BrandonLim8890, line 18: So this is giving me some warnings upon starting the app, but its required for class-validator to work with react hook form. If anyone knows the reason to this, I'd be interested in understanding why these warnings are popping up!

packages/common/src/custom-validators.ts
```diff
@@ -0,0 +1,29 @@
+import { 
+  registerDecorator, 
+  ValidationArguments, 
+  ValidationOptions 
+} from 'class-validator';
+
+export function IsEqualTo(property: string, validationOptions?: ValidationOptions) {
```
BrandonLim8890, line 7: This is a custom validator for class-validator. This will essentially ensure that two values in a dto matches.

packages/frontend-v2/temp/dto-types.ts
```diff
@@ -0,0 +1,218 @@
+import {
```
BrandonLim8890, line 1: Class-validator and decorators are being weird when importing from `@graduate/common`. To sidestep this blocker, I created a temporary `dto-types` file in the frontend-v2 package so that I can ensure my code works. Fixing class-validator and decorators in the common package will be another ticket.

babel.config.js
bugsalexander: i think its fine. i remember doing something along these lines for KOH https://github.com/sandboxnu/office-hours/blob/master/packages/app/babel.config.js

package.json
```diff
@@ -25,6 +25,7 @@
     "postinstall": "husky install"
   },
   "dependencies": {
+    "@babel/plugin-proposal-class-properties": "^7.18.6",
```
bugsalexander, line 28: RE: to the above tho, we should move to using next to package stuff and api to package stuff themselves asap cuz maintaining babel is a PITA

packages/common/src/dto-types.ts
```diff
@@ -76,16 +77,26 @@ export class UpdatePlanDto {
...
-}
 
+  @IsNotEmpty({ message: "Confirm Password cannot be empty" })
+  @IsString({
+    message: "Confirm Password must only contain alphanumeric values and signs",
+  })
+  @IsEqualTo("password", {
+    message: "Confirm Password must be same as Password",
+  })
+  passwordConfirm: string;
```
bugsalexander, line 98: sorry if u put a lot of work into the decorator thing, but i think more complex validation (that is not essentially jsut property checking) should be put in the actual api

packages/frontend-v2/package.json
```diff
@@ -14,11 +14,14 @@
     "@emotion/styled": "^11",
     "@fontsource/montserrat-alternates": "^4.5.9",
     "@graduate/common": "workspace:*",
+    "@hookform/resolvers": "^2.9.3",
+    "class-validator": "^0.13.2",
```
bugsalexander, line 18: why is class-validator needed on the frontend? validation should (probably) be backend-only, unless it's just field inputs (in which case we use something else, idrmb what it is)

packages/frontend-v2/pages/login.tsx
```diff
@@ -0,0 +1,135 @@
...
+      if (user) {
+        if (user.isOnboarded) {
+          // redirect home
+          console.log("redirect to home");
+          router.push("/home");
+        } else {
+          // redirect to onboarding
+          console.log("redirect to onboarding");
+          router.push("/onboarding");
+        }
```
bugsalexander, line 69: this code looks similar to earlier code - can pullout and abstract

packages/frontend-v2/pages/login.tsx
```diff
@@ -0,0 +1,135 @@
...
+      }
+    } catch (err) {
+      const error = err as AxiosError;
+      if (error.response?.status === 401)
+        setApiError("Invalid credentials, please try again.");
+      else setApiError(error.message);
+      console.log(error);
```
bugsalexander, line 76: instead of console.log, use logger.error or something

packages/frontend-v2/pages/login.tsx
```diff
@@ -0,0 +1,135 @@
...
+  const [apiError, setApiError] = useState("");
+  const [renderSpinner, setRenderSpinner] = useState(true);
+  const router = useRouter();
+
+  const {
+    register,
+    handleSubmit,
+    formState: { errors, isSubmitting },
+  } = useForm<LoginStudentDto>({
+    resolver: classValidatorResolver(LoginStudentDto),
```
bugsalexander, line 28: what does this do? not sure how much validation you really want to do here, because most of it should be handled via api errors anyway

packages/frontend-v2/pages/login.tsx
```diff
@@ -0,0 +1,135 @@
...
+      <FormControl isInvalid={errors.password != null}>
+        <InputGroup>
+          <Input
+            type={"password"}
+            id="password"
+            placeholder="Enter Password"
+            {...register("password")}
+          />
+        </InputGroup>
+        <FormErrorMessage>{errors.password?.message}</FormErrorMessage>
```
bugsalexander, line 120: how can i manually test the frontend? i go to 127.0.0.1:3002 and can't seem to click the sign in button (it does nothing). then also visiting /login results in an error see belo
Screen Shot 2022-07-29 at 8 53 53 PM
w

packages/frontend-v2/pages/signup.tsx
```diff
@@ -0,0 +1,115 @@
...
+  const [apiError, setApiError] = useState("");
+
+  const router = useRouter();
+
+  const {
+    register,
+    handleSubmit,
+    formState: { errors, isSubmitting },
+  } = useForm<SignUpStudentDto>({
+    resolver: classValidatorResolver(SignUpStudentDto),
```
bugsalexander, line 28: same comment as above, probably save complex validtion for server side

packages/common/src/dto-types.ts
BrandonLim8890: I'm not sure if I understand this correctly, but NestJS uses class-validator to perform its body validation. So our API actually uses the properties and decorators of class-validator to perform validation. Do you mean we instead switch to having the auth controller directly check if `confirmPassword == password`?

packages/frontend-v2/package.json
BrandonLim8890: Cuz I was having errors with importing the dto-types into the frontend, I made a mock dto-types.ts file in the frontend temp folder. After we fix the importing dto-types error, we can remove this.

packages/frontend-v2/pages/login.tsx
BrandonLim8890: This will basically perform all the same validation that the backend has on the frontend. It ensures that the email is an email, and that the password provided is a string

BrandonLim8890:
[Login Page](https://trello.com/c/sdrPIzCb/437-login-page)

BrandonLim8890:
[Sign up Page](https://trello.com/c/mygjhx0k/436-sign-up-page)

packages/frontend-v2/pages/login.tsx
aryanshah701: since we can reuse the dto classes with the class-validator decorators to do the same validation on the frontend that the backend does I think it's fine to validate on the frontend, saves us an api call for data that is obviously malformed without any additional logic needed

packages/frontend-v2/pages/signup.tsx
aryanshah701: the dtos don't necessarily do complex validation? they mostly just validate the shape of the json with some sanity constraints so i think it may be fine to do the validation client side especially since we don't need any additional logic

bugsalexander: will the backend not need to re-do the validation anyway? i think pulling in an extra library and dependency is not worth it

packages/common/src/dto-types.ts
bugsalexander: yes, that's what i meant

bugsalexander: i think for simple validation like typechecking and the builtins that allow us to do value checks, it's good to have nest do the heavy lifting. but this specific annotation will probably never be used in another situation, so it's not worth the extra complexity of abstracting it

packages/frontend-v2/pages/login.tsx
bugsalexander: you can use input type="email" to do email validation. and the password that is provided will always be a string, from the input. if it isn't, the UI should handle it

packages/frontend-v2/pages/login.tsx
bugsalexander: @BrandonLim8890 are you able to test it locally?

BrandonLim8890: Yea I can test locally. I'm not sure why it's giving you that error

packages/frontend-v2/package.json
bugsalexander: just so you don't forget anything, can you leave a list of TODOs somewhere?

packages/frontend-v2/pages/index.tsx
```diff
@@ -25,25 +30,28 @@ interface InfoTextProps {
...
-      <Banner />
+      <Header router={router} />
+      <Banner router={router} />
       <Info />
     </Box>
   );
 };
 
-const Header = (): JSX.Element => {
+const Header = ({ router }: RouterProps): JSX.Element => {
```
bugsalexander, line 43: in this situation (see next comment) you don't need router. however if you did, you can just access the router directly in this component, it doesn't need to be a prop

packages/frontend-v2/pages/index.tsx
```diff
@@ -25,25 +30,28 @@ interface InfoTextProps {
...
   return (
     <HeaderContainer>
       <Logo />
-      <Button size="sm">Sign In</Button>
+      <Button size="sm" onClick={() => router.push("/login")}>
```
bugsalexander, line 47: you should use next/link for this instead of router

packages/frontend-v2/pages/index.tsx
```diff
@@ -84,6 +92,7 @@ const Banner = (): JSX.Element => {
           <Button
             mr={{ desktop: "7.5rem", laptop: "6.25rem", tablet: "3.25rem" }}
             mt="15%"
+            onClick={() => router.push("/signup")}
```
bugsalexander, line 95: use next/link here

packages/frontend-v2/pages/login.tsx
aryanshah701: fair fair, i'm okay with not using class-validator on the frontend and keeping validation simple as alex mentioned

packages/frontend-v2/pages/signup.tsx
aryanshah701: fair, good with having class validator remove from the frontend

packages/frontend-v2/utils/apiRouting.ts
```diff
@@ -0,0 +1,19 @@
+import { GetStudentResponse } from "@graduate/common";
+import { NextRouter } from "next/router";
+
+export const routeOnboarding = (
```
BrandonLim8890, line 4: Not sure if this name/file name is correct. Any suggestions for what I should rename this/where I should place this?

packages/frontend-v2/pages/login.tsx
```diff
@@ -0,0 +1,131 @@
...
+    } catch (err) {
+      const error = err as AxiosError;
+      logger.error(error);
+      setRenderSpinner(false);
+    }
+  };
+
+  useEffect(() => {
+    loginWithCookie();
+  }, []);
```
aryanshah701, line 53: good use case for a custom hook "useRedirectIfLoggedIn()" that you can just call at the top of this component, it'll just clean up the component and can probably be reused in sign up too

aryanshah701 (approved):
Looks good to me, awesomeeee workk!! Just a few nits and then feel free to merge :)

packages/frontend-v2/pages/login.tsx
BrandonLim8890: Hmm, I can make this a custom hook but I dont think it would be smart to do this for Signup from a UX perspective. If a user logs out and wants to create a new user but still has the JWT in cookies, then we wouldn't want to redirect them to their old page if they're trying to sign up as a new user. What are your thoughts?

aryanshah701: sounds good!

PR #438, opened 2022-08-10
Add basic logging to api-v2

aryanshah701:
# Description

A basic setup of logging for Graduate's backend. We do 2 types of logging as of now
- Log every request, response, and error using an interceptor
- Service debug logs for instances where we need to disambiguate ambiguous nulls returned by the service

We also have a basic setup of filtering out blacklisted keys from all log messages, this will prevent us from logging sensitive information.

This setup is basic and helps us get started with logging, as we scale we shall mature our logging.

Closes [Add logging to backend](https://trello.com/c/G3yi5qhq/455-add-logging-to-backend)

## Type of change

# How Has This Been Tested?

- [x] Manually tested by calling the API through postman and viewing the logs in the console 
- [x] Basic unit tests for deepFilter function

### Create user success - note the interceptor logs and how password field is not being logged since it is blacklisted
Screen Shot 2022-08-10 at 10 50 55 AM

### Create user error - note the interceptor logs and service debug log
Screen Shot 2022-08-10 at 10 49 35 AM
Screen Shot 2022-08-10 at 10 49 54 AM

# Checklist:


packages/api-v2/src/plan/plan.service.ts
```diff
@@ -18,8 +21,7 @@ export class PlanService {
 
     try {
       return this.planRepository.save(newPlan);
-    } catch (_error) {
-      // TODO: Add logging for errors here and in every service method.
+    } catch (error) {
       return null;
```
rael346, line 25: Just curious but why don't we log the error here?

packages/api-v2/src/auth/auth.service.ts
```diff
@@ -16,9 +19,7 @@ export class AuthService {
     private readonly jwtService: JwtService
   ) {}
 
-  /**
-   * Registers a new student in the db and logs the student in.
-   */
+  /** Registers a new student in the db and logs the student in. */
   async register(createStudentDto: CreateStudentDto): Promise<Student> {
     // create a new student
     const newStudent = await this.studentService.create(createStudentDto);
```
rael346, line 25: If `newStudent` is null here, shouldn't we log this instead of just returning `null`?

packages/api-v2/package.json
```diff
@@ -29,7 +29,7 @@
     "@nestjs/passport": "^8.1.0",
     "@nestjs/platform-express": "^8.0.0",
     "@nestjs/typeorm": "^8.0.2",
-    "bcrypt": "^5.0.1",
+    "bcrypt": "5.0.0",
```
rael346, line 32: Maybe add a comment noting why we downgrade `bcrypt`? (5.0.1 not working with M1, giving me this error `mach-o file, but is an incompatible architecture (have 'x86_64', need 'arm64e')`)

rael346 (approved):
LGTM

packages/api-v2/src/auth/auth.service.ts
aryanshah701: ahh yea, i didn't log here cause we don't know the reason newStudent is null here... we only know the reason in studentService.create which in turn does log. If we logged here all we could do is say that create returned null for some reason which isn't very insightful

packages/api-v2/package.json
aryanshah701: ahh yea, adding comments is weird with package.json... but we should be good if people reviewing this pr see your comment here for some context

.prettierrc
```diff
@@ -4,7 +4,7 @@
   "semi": true,
   "tabWidth": 2,
   "printWidth": 80,
-  "plugins": ["prettier-plugin-jsdoc"],
+  "plugins": ["./node_modules/prettier-plugin-jsdoc"],
```
bugsalexander, line 7: thanks for fixing this!

packages/api-v2/graduate-logger.ts
```diff
@@ -0,0 +1,11 @@
+import { ConsoleLogger, LogLevel } from "@nestjs/common";
+import { deepFilter } from "src/utils";
+
+const BLACKLIST = ["password"];
```
bugsalexander, line 4: replace with inclusive language – "denylist"

packages/api-v2/src/interceptors/logging.interceptor.ts
```diff
@@ -0,0 +1,144 @@
...
+    const { method, url } = req;
+    const { statusCode } = res;
+    const ctx = this.formatCtx(LogType.Response, method, url, statusCode);
+
+    this.logger.log(
+      {
+        body,
+      },
+      ctx
+    );
```
bugsalexander, line 83: did the autoformatter do this or can you do

```javascript
this.logger.log({ body }, ctx);
```

bugsalexander (commented):
did a SUPER cursory review, didn't look at functionality at all just code practices

PR #441, opened 2022-08-19
Frontend V2 - Basic Plan Component

aryanshah701:
Update: Something went wrong and github isn't able to properly create the diff against the previous Plan PR[fetch plans and populate dnd ids]. Instead of spending time to figure it out I just closed the previous PR and this will have all the changes to avoid any confusion. Sorry :( 

Note: All PRs for plan after this one will be "stacked" on top of one another. This will allow me to create smaller PRs quickly and also make it easier for yall to see smaller diffs. Each PR will show the diff against the previous stage of the Plan component stages. For more details on the stages I plan to be implementing this component in, look below.

# Description

This PR adds functionality to display a student's primary plan with courses, drag courses between terms and post the updated plan to the backend.

An overview of all the functionality completed and remaining for the plan component(note, there may be more stuff remaining but this is what i can think of for now).

- [x] fetch plans and populate dnd ids: current PR
- [x] basic plan component with dnd and posting to API: current PR
- [ ] styling
- [ ] choosing which plan to display
- [ ] choosing primary plan
  - [ ] fix backend bug of allowing users to choose a primary plan that doesn't belong to this(massive security issue) 
- [ ] validation and errors
- [ ] add class validator decorators to schedule type so that we enforce at least some shape of the schedule when posting
- [ ] figure out how to show "preview" of course in the position it was when it is being dragged around
- [ ] batch/throttle plan updates

## Type of change

# How Has This Been Tested?

- Currently just been manually tested. 
  - To manually test, create a new user with creds "aryan@gmail.com" and "aryan1234".
  - Create a plan (add at least a few years with a couple of semesters and courses, refer to the Schedule2 type in the [types file in common](https://github.com/sandboxnu/graduatenu/blob/aryan/v2-plan-fetching/packages/common/src/types.ts) to understand the shape of the schedule
  - Update student to set the primary plan id as the plan that was just created
  - Visit frontend, click on sign in, and play around with the plan
- Working on adding some basic unit tests for the involved functions.

I'll add a video for this soon :))

# Checklist:


packages/api-v2/src/guards/own-plan.guard.ts
```diff
@@ -54,5 +54,7 @@ export class OwnPlanGuard implements CanActivate {
         formatServiceCtx("OwnPlanGuard", "canActivate")
       );
     }
+
+    return res;
```
aryanshah701, line 58: bug from logging pr, oops

packages/api-v2/src/student/student.controller.ts
```diff
@@ -45,6 +45,10 @@ export class StudentController {
       throw new BadRequestException();
     }
 
+    if (!isWithPlans) {
+      student.plans = [];
+    }
```
aryanshah701, line 50: makes it much easier to deal with an empty array than potentially undef on the frontend

packages/frontend-v2/hooks/useStudentWithPlans.ts
```diff
@@ -54,72 +32,17 @@ export function useStudentWithPlans(): UseStudentReturn {
...
-      id: `${course.classId}-${course.subject}-${updatedCount}`,
-    };
-  });
-
-  const dndTerm = {
-    ...term,
-    id: `${term.year}-${term.season}`, // add a unique dnd id to all terms
-    classes: dndClasses,
-  };
-
```
aryanshah701, line 120: moved functions to utils

packages/frontend-v2/hooks/useStudentWithPlans.ts
```diff
@@ -28,23 +22,7 @@ type UseStudentReturn = StudentResponse & {
...
-      }
-
-      // prepare all of the student's plans for drag and drop by adding drag and drop ids
-      const plansWithDndIds = student.plans.map(preparePlanForDnd);
-      return {
-        ...student,
-        plans: plansWithDndIds,
-      };
-    }
-  );
```
aryanshah701, line 47: moved into its own function to reuse in home

packages/frontend-v2/pages/index.tsx
```diff
@@ -35,10 +37,17 @@ const Home: NextPage = () => {
 };
 
 const Header = (): JSX.Element => {
+  const router = useRouter();
+  const signUp = async () => {
+    await API.auth.login({ email: "aryan@gmail.com", password: "aryan1234" });
+    router.push("/home");
+  };
```
aryanshah701, line 44: work around till login gets merged

packages/frontend-v2/hooks/useStudentWithPlans.ts
```diff
@@ -29,25 +22,7 @@ type UseStudentReturn = StudentResponse & {
...
```
aryanshah701, line 50: moved into its own function for reusability in home

packages/frontend-v2/hooks/useStudentWithPlans.ts
```diff
@@ -58,92 +33,16 @@ export function useStudentWithPlans(): UseStudentReturn {
...
-      id: `${course.classId}-${course.subject}-${updatedCount}`,
-    };
-  });
-
-  const dndTerm = {
-    ...term,
-    id: `${term.year}-${term.season}`, // add a unique dnd id to all terms
-    classes: dndClasses,
-  };
-
```
aryanshah701, line 144: moved to utils

packages/frontend-v2/utils/plan/preparePlanForDnd.ts
```diff
@@ -0,0 +1,97 @@
+import {
```
aryanshah701, line 1: functions moved from useStudentWithPlan, can ignore if you https://github.com/sandboxnu/graduatenu/pull/439 was reviewed.

packages/common/src/types.ts
```diff
@@ -353,10 +381,59 @@ export interface RANGESection {
...
+  summer2: ScheduleTerm2<T>;
+  isSummerFull: boolean;
+}
+
+/**
+ * A clean version of the ScheduleTerm used by V2 of the App. A generic id field
+ * is used for book keeping purposes by the drag and drop library, in cases
+ * where we don't care about this id, T can null.
+ */
+export interface ScheduleTerm2<T> {
```
BrandonLim8890, line 405: A little confused on why we need a generic type

aryanshah701: good point, we could get away with the id field being string | null. alex's design had generic to allow the type to be more flexible, in case we want the book keeping id field to be more complicate(maybe an object w multiple fields).

i think string | null is a good choice and while it isn't fully clear as to which option is better, both options work and aren't too bad so imma leave it as is

packages/frontend-v2/pages/home.tsx
```diff
@@ -0,0 +1,128 @@
...
+    // TODO: Handle no plan/no primary plan case
+    throw new Error("Student doesn't have a valid primary plan set");
+  }
+
+  const handleDragStart = (event: DragStartEvent): void => {
+    // find the course that is being dragged and set it as the active course
+    const { active } = event;
+    const activeCourse = findCourseByDndId(
+      primaryPlan.schedule,
+      active.id as string // dnd ids are strings
```
BrandonLim8890, line 56: If im understanding correctly, this type has to match the generic for a PlanModel<T>. If this is the case, should make a note that this type must match that generic in case we want to change the type of the id for whatever reason

packages/frontend-v2/pages/home.tsx
```diff
@@ -0,0 +1,128 @@
...
+      // update failed, either due to some logical issue or explicitely thrown error
+      logger.debug("updatePlanOnDragEnd", err);
+      return;
+    }
+
+    // create a new student object with this updated plan so that we can do an optimistic update till the API returns
+    const updatedStudent = updatePlanForStudent(student, updatedPlan);
+
+    // post the new plan and update the cache
+    mutateStudent(
```
BrandonLim8890, line 88: I LOVE SWR!

packages/frontend-v2/hooks/useStudentWithPlans.ts
```diff
@@ -10,8 +11,8 @@ type StudentResponse = Omit<
 
 type UseStudentReturn = StudentResponse & {
   isLoading: boolean;
-  mutateStudent: KeyedMutator<GetStudentResponse>;
-  student?: GetStudentResponse;
+  mutateStudent: KeyedMutator<StudentModel<string>>;
```
BrandonLim8890, line 14: Is this changed because we need the id type of string for each course when updating the plan?

BrandonLim8890 (approved):
Overall LGTM! Very easy to read and understand after reading dnd docs

packages/frontend-v2/pages/home.tsx
aryanshah701: AHH good point!!

packages/frontend-v2/pages/home.tsx
aryanshah701: ME TOOOO

packages/frontend-v2/hooks/useStudentWithPlans.ts
aryanshah701: yeaaa, it's cause we don't return the student but rather a processed students that has their plans populated with dnd ids

PR #445, opened 2022-08-26
Frontend V2 - Switch Plans

aryanshah701:
# Description

https://user-images.githubusercontent.com/30478978/186790661-a22e5571-ad69-45bf-bcd3-a1e6c8170b63.mov

Reminder: This PR is "stacked" on top of v2-plan-toggle-year. This diff will only show changes specific to switching plans.

Allows the user to switch between multiple plans. Displays primary plan if the user has one.

An overview of all the functionality completed and remaining for the plan component(note, there may be more stuff remaining but this is what i can think of for now).

- [x] fetch plans and populate dnd ids
- [x] basic plan component with dnd and posting to API
- [x] styling
- [x] collapsing years: previous PR
- [x] **choosing which plan to display** current PR
- [ ] adding a course using the add course modal
- [ ] adding a Plan
- [ ] choosing primary plan
  - [ ] fix backend bug of allowing users to choose a primary plan that doesn't belong to this(massive security issue) 
- [ ] validation and errors
- [ ] add class validator decorators to schedule type so that we enforce at least some shape of the schedule when posting
- [ ] figure out how to show "preview" of course in the position it was when it is being dragged around
- [ ] batch/throttle plan updates

## Type of change

# How Has This Been Tested?

Manually. See video. 

# Checklist:

### DIFF
--- a/packages/frontend-v2/components/Plan/PlanDropdown.tsx
+++ b/packages/frontend-v2/components/Plan/PlanDropdown.tsx
@@ -3,9 +3,9 @@ import { Select } from "@chakra-ui/react";
 import { PlanModel } from "@graduate/common";
 
 interface PlanDropdownProps {
-  setSelectedPlanId: Dispatch<SetStateAction<number | undefined>>;
+  setSelectedPlanId: Dispatch<SetStateAction<number | undefined | null>>;
   plans: PlanModel<string>[];
-  selectedPlanId?: number;
+  selectedPlanId: number | undefined | null;
 }
 export const PlanDropdown: React.FC<PlanDropdownProps> = ({
   setSelectedPlanId,
@@ -18,8 +18,14 @@ export const PlanDropdown: React.FC<PlanDropdownProps> = ({
       width="15%"
       mb="sm"
       borderRadius="0"
-      value={selectedPlanId}
+      value={selectedPlanId ? selectedPlanId : undefined}
       onChange={(e) => {
+        if (!e.target.value) {
+          // no plan is selected, indicated using null(different from undef which is the initial state)
+          setSelectedPlanId(null);
+          return;
+        }
+
         const selectedPlanId = parseInt(e.target.value);
         setSelectedPlanId(selectedPlanId);
       }}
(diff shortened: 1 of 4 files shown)


packages/frontend-v2/components/Plan/PlanDropdown.tsx
```diff
@@ -0,0 +1,40 @@
...
+    <Select
+      placeholder="Select Plan"
+      width="15%"
+      mb="sm"
+      borderRadius="0"
+      value={selectedPlanId ? selectedPlanId : undefined}
+      onChange={(e) => {
+        if (!e.target.value) {
+          // no plan is selected, indicated using null(different from undef which is the initial state)
+          setSelectedPlanId(null);
```
BrandonLim8890, line 25: is there a reason why we don't use undef instead of null here? what is the importance of differentiating between unselected, vs no plan. Idk if its good to have a situation where there is just no plan in the home page

packages/frontend-v2/pages/home.tsx
```diff
@@ -18,46 +24,74 @@ import { PlanModel } from "@graduate/common";
...
+   * 1. Create a copy of the plan with the course remove from the old schedule and
+   *    place into the new schedule
+   * 2. Create a copy of the student with the new updated plan
+   * 3. Optimistically update the local cache with the new student so that the user
+   *    sees the change immidiately
+   * 4. POST the updated plan to db
+   * 5. Refetch the student with the new plan and update the cache with the
+   *    persisted student from our backend
+   * 6. If anything goes wrong in the POST, rollback the optimistic update
+   */
```
BrandonLim8890, line 77: I love this

BrandonLim8890 (approved):
LGTM! 
Also side design note: While using the UI, I feel like it could be a lot more comfortable if there was also an expand all/collapse all years button.

packages/frontend-v2/components/Plan/PlanDropdown.tsx
aryanshah701: You're right! The unselected is when the user selects the "no selection" option in the dropdown. Now it makes sense to always display a plan but dropdown menus normally allow you to have the "no selection" option. Shall discuss this with designers.

PR #469, opened 2022-09-26
Frontend V2 - Adding A New Plan

aryanshah701:
# Description

We can now create a new plan.

Callouts
- The UX for this feature isn't finalized so I just made it look decent for now. 
- We also don't have majors/concentrations/catalog year data scraped yet so going off of some mock data for now. 
- This stuff is mostly used for requirements validation so we'll fix it when we implement that.

Other changes
- Remove the coop cycle field from a Plan
- Fix all imports from utility files to import from utils directory 

Closes https://github.com/sandboxnu/graduatenu/projects/1#card-85686954

## Type of change

# How Has This Been Tested?

Manually tested by adding a new plan and playing around with it.

# Checklist:

- [x I have performed a self-review of my own code

### DIFF
--- a/packages/api-v2/src/plan/entities/plan.entity.ts
+++ b/packages/api-v2/src/plan/entities/plan.entity.ts
@@ -28,9 +28,6 @@ export class Plan {
   @Column()
   major: string;
 
-  @Column()
-  coopCycle: string;
-
   @Column({ nullable: true })
   concentration: string;
 
--- a/packages/frontend-v2/components/AddCourseModal/SearchResult.tsx
+++ b/packages/frontend-v2/components/AddCourseModal/SearchResult.tsx
@@ -1,7 +1,7 @@
 import { SmallAddIcon } from "@chakra-ui/icons";
 import { Flex, Box, Heading, IconButton, Text } from "@chakra-ui/react";
 import { ScheduleCourse2 } from "@graduate/common";
-import { getCourseDisplayString } from "../../utils/course/getCourseDisplayString";
+import { getCourseDisplayString } from "../../utils/";
 import { GraduateToolTip } from "../GraduateTooltip";
 
 interface SearchResultProps {
--- a/packages/frontend-v2/components/Plan/AddPlanModal.tsx
+++ b/packages/frontend-v2/components/Plan/AddPlanModal.tsx
@@ -1,0 +1,171 @@
+import {
+  Button,
+  Flex,
+  FormControl,
+  FormErrorMessage,
+  FormLabel,
+  Input,
+  Modal,
+  ModalBody,
+  ModalCloseButton,
+  ModalContent,
+  ModalFooter,
+  ModalHeader,
+  ModalOverlay,
+  Select,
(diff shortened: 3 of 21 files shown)


packages/api-v2/src/plan/entities/plan.entity.ts
```diff
@@ -28,9 +28,6 @@ export class Plan {
   @Column()
   major: string;
 
-  @Column()
-  coopCycle: string;
-
```
aryanshah701, line 33: we don't need this field anymore cause it is quite useless, there are a bunch of changes regarding this

packages/frontend-v2/components/AddCourseModal/SearchResult.tsx
```diff
@@ -1,7 +1,7 @@
 import { SmallAddIcon } from "@chakra-ui/icons";
 import { Flex, Box, Heading, IconButton, Text } from "@chakra-ui/react";
 import { ScheduleCourse2 } from "@graduate/common";
-import { getCourseDisplayString } from "../../utils/course/getCourseDisplayString";
```
aryanshah701, line 4: a couple of changes having to do with importing directly from "utils" rather than the actual file

BrandonLim8890 (approved):
LGTM! Overall very clean and easy to follow. Only 1 small suggestion I made. 
An additional thing, I tried playing around with the `close()` function for the modal because I saw that the form clears just before closing, which I think from a UI/UX perspective looks a little funny? If its possible, I think it would be better for the form to only clear once the Modal has fully closed so that users don't think that the form was cleared before submitting or smt like that. I think it just looks cleaner and more professional to have the form values persist until the modal is closed. See attached for what I mean

testmodal

packages/frontend-v2/components/Plan/AddPlanModal.tsx
```diff
@@ -0,0 +1,170 @@
...
+              </FormControl>
+            </ModalBody>
+            <ModalFooter>
+              <Button isLoading={isSubmitting} type="submit" ml="auto">
+                Add Plan
+              </Button>
+            </ModalFooter>
+          </form>
+        </ModalContent>
+      </Modal>
```
AlpacaFur, line 155: This seems like it might be a lot to fit in one component from a lines-of-code/readability persepective. It is all still clearly part of one unit though, so it might be okay as is.

AlpacaFur (approved):
Made one comment, but other than that it looks good!

packages/frontend-v2/components/Plan/AddPlanModal.tsx
aryanshah701: I totally agree, i think the only thing making me a little less inclined in doing so is that there isn't any code repeatability that we'd eliminate by separating into multiple reusable components. On the other hand, separating this into multiple presentational components will require drilling props and a lot of additional code.

Honestly, i think i may be a little lazy but since there is somewhat of a tradeoff i shall let it be haha, regardless a good point!!

PR #470, opened 2022-09-27
Styling for Login and Signup Page

### LINKED ISSUE #450: Style Login and Signup Pages
Style the login and signup pages according to the designs on [figma](https://www.figma.com/file/yUflVFVxjMynm7gGymLpQY/%F0%9F%8F%97-Current-Work?node-id=1831%3A8223).

BrandonLim8890:
# Description

Add styling to login and signup pages. Very simple styling with ChakraUI and I also added more options to the themes, as well as additional reusable components.

login-styling

Closes #450 

## Type of change

# How Has This Been Tested?

Played around with the rest of the UI as well as login/signup flow and everything works

# Checklist:


packages/frontend-v2/components/Form/Input.tsx
```diff
@@ -0,0 +1,33 @@
...
+import { forwardRef } from "react";
+
+interface InputProps {
+  error: FieldError | undefined;
+  id: string;
+  placeholder: string;
+  type: string;
+}
+
+// eslint-disable-next-line react/display-name
```
BrandonLim8890, line 17: forward refs are weird with this. Not sure how to bypass without adding this line

packages/frontend-v2/components/Button/FormButtons.tsx
```diff
@@ -0,0 +1,35 @@
+import { Button, ButtonProps, ComponentWithAs } from "@chakra-ui/react";
+
+export const SubmitButton: ComponentWithAs<"button", ButtonProps> = ({
+  children,
```
rael346, line 4: Is there a reason for not using `forwardRef` here but use it for the `StringInput`?

packages/frontend-v2/pages/signup.tsx
```diff
@@ -37,81 +70,97 @@ const Signup: NextPage = () => {
...
+    <InputGroup>
+      <StringInput
+        id="email"
+        placeholder="Email"
+        error={errors.email}
+        type="email"
+        {...register("email", {
+          required: "Email is required",
+          pattern: {
+            value: /^[A-Z0-9._%+-]+@[A-Z0-9.-]+\.[A-Z]{2,}$/i,
```
rael346, line 115: It is probably better to use a schema validation library like `yup` or `zod` to check for correct input type here since using regex is prone to errors and not sustainable.

rael346 (approved):
LGTM!

packages/frontend-v2/pages/signup.tsx
BrandonLim8890: We discussed using this earlier in the logic portion but decided against since it didn't make sense to complicate the frontend validation

packages/frontend-v2/components/Button/FormButtons.tsx
BrandonLim8890: I think its general consensus to avoid using forward refs whenever possible, cuz this is just a component that calls a parent's function

PR #473, opened 2022-10-13
Add support for deleting a plan

### LINKED ISSUE #453: Add Support To Delete An Existing Plan

aryanshah701:
# Description

https://user-images.githubusercontent.com/30478978/195732696-d04f0c89-79ef-40c7-a1d8-91ef6dcb2b03.mov

Support to delete a plan. UX is jank and ad-hoc, will go in an update it once designers finalize the UX for the feature.

Closes #453

## Type of change

# How Has This Been Tested?

Manually, see video.

# Checklist:

### DIFF
--- a/packages/frontend-v2/components/Plan/DeletePlanModal.tsx
+++ b/packages/frontend-v2/components/Plan/DeletePlanModal.tsx
@@ -1,0 +1,60 @@
+import {
+  Button,
+  Modal,
+  ModalBody,
+  ModalCloseButton,
+  ModalContent,
+  ModalFooter,
+  ModalHeader,
+  ModalOverlay,
+} from "@chakra-ui/react";
+import { API } from "@graduate/api-client";
+import { useRouter } from "next/router";
+import { mutate } from "swr";
+import { USE_STUDENT_WITH_PLANS_SWR_KEY } from "../../hooks";
+import { handleApiClientError } from "../../utils";
+import { GrayButton } from "../Button";
+
+interface DeletePlanModalProps {
+  planName: string;
+  planId: number;
+  isOpen: boolean;
+  onClose: (isDeleted: boolean) => void;
+}
+export const DeletePlanModal: React.FC<DeletePlanModalProps> = ({
+  planName,
+  planId,
+  isOpen,
+  onClose,
+}) => {
+  const router = useRouter();
+  const closeWithoutDelete = () => onClose(false);
+  const deletePlan = async () => {
+    try {
+      await API.plans.delete(planId);
+
+      // refresh the cache and close the modal
+      mutate(USE_STUDENT_WITH_PLANS_SWR_KEY);
+      onClose(true);
+    } catch (error) {
+      handleApiClientError(error as Error, router);
+    }
+  };
+
+  return (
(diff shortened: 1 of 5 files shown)


packages/frontend-v2/components/Plan/DeletePlanModal.tsx
```diff
@@ -0,0 +1,60 @@
...
+      <ModalOverlay />
+      <ModalContent>
+        <ModalHeader>Delete Plan</ModalHeader>
+        <ModalCloseButton />
+        <ModalBody>Woah, are you sure you want to delete {planName}?</ModalBody>
+        <ModalFooter>
+          <Button mr={3} onClick={deletePlan}>
+            Delete
+          </Button>
+          <GrayButton onClick={closeWithoutDelete}>Nevermind</GrayButton>
```
rael346, line 55: I think this `GrayButton` is unnecessary since we already have `ModalCloseButton` above

packages/frontend-v2/pages/home.tsx
```diff
@@ -175,6 +206,43 @@ const HomePage: NextPage = () => {
...
+interface DeletePlanButtonProps {
+  onOpen: () => void;
+  onClose: (isDeleted: boolean) => void;
+  isOpen: boolean;
+  planName: string;
+  planId: number;
+}
+
+const DeletePlanButton: React.FC<DeletePlanButtonProps> = ({
+  onOpen,
```
rael346, line 218: I think we should move this button (and the disclosure hooks and logics) to the `DeletePlan` component, so that it encapsulates the modal logic better. The new modal will look something like
```Typescript
type DeletePlanModalProps = {
   planName: string;
   planId: number;
}

const DeletePlanModal: React.FC<DeletePlanModalProps>({
   planName,
   planId,
}) => {
   const { onOpen, onClose, isOpen } = useDisclosure();
   
   return (
        <>
             <Button/>
             <Modal>
                    ...
             <Modal/>
        </>
   )
}
```

This will remove the similar type (`DeletePlanButtonProps` and `DeletePlanModalProps `), remove the `useDisclosure` hook from cluttering the `Home` page (and also remove the conflicting name from the modals), and the `onOpen`, `onClose` and `isOpen` parameters from the modal component.

rael346 (commented):
Left some comments. Overall LGTM!

packages/frontend-v2/pages/home.tsx
jia614: agree that it makes more sense to have the delete plan button live in the components directory and be imported into home.tsx-- it'd be nice to do something similar for the add modal button as well! :)

jia614 (approved):
just a small refactor request, but otherwise lgtmeeeee 👍

packages/frontend-v2/components/Plan/DeletePlanModal.tsx
aryanshah701: Trueee, but most delete confirmations one have a large button next to the confirm delete button to abort so that people don't delete by mistake. regardless we'll have actual designs for the modal so we can update it based on that :))

packages/frontend-v2/pages/home.tsx
aryanshah701: good point, thank guys!

PR #474, opened 2022-10-14
Add Major Requirement Sidebar

### LINKED ISSUE #456: Implement Sidebar That Displays Major Requirements With Draggable Courses

AlpacaFur:
# Description

Implements sidebar + basic styling that allows required courses to be dragged into the schedule. General additional styling fixes and home page cleanup. 

This PR is primarily meant to show all of a given major's requirements in the sidebar, call Search's API to get course info for each requirement and populate the draggable course, and enable drag and drop from the sidebar into the plan. It **does not yet** validate the plan against the sidebar. 

The way we're fetching data from Search right now (requesting per course) does not work well (takes a long time and sometimes not every course resolves). Search has graciously created a batch endpoint that allows us to fetch multiple courses at once, so we will update our API call and implementation accordingly as a performance improvement for the sidebar.

Future work (in PRs) for the sidebar: 
- Optimizing the sidebar (utilizing Search's new endpoint to fetch courses more quickly and more consistently)
- Fixing drag and drop lag (dragging courses in the home component is currently very slow when they are all rendered)
- Validation

https://user-images.githubusercontent.com/2746893/195738221-efc9c04f-df96-403e-903d-710248131f33.mov

Closes #456 

## Type of change

# How Has This Been Tested?

- [x] We tested manually by checking that courses are draggable, update the schedule in the database, and load in from Search.

# Checklist:


packages/frontend-v2/components/Sidebar/Sidebar.tsx
```diff
@@ -0,0 +1,292 @@
...
+      },
+      []
+    );
+    // console.info("Making", requirements.length, "requests")
+
+    const promises: Promise<ScheduleCourse2<null> | null>[] = [];
+
+    for (const requirement of requirements) {
+      promises.push(
+        // new Promise((res, rej) => {
```
jia614, line 73: this commented stuff mimics the fetch course api client calls since ours are currently scuffed :(. you can use this segment to test with "fully loaded" courses in the sidebar

packages/frontend-v2/utils/convertMajor.ts
```diff
@@ -0,0 +1,303 @@
+/**
```
jia614, line 1: this is a utils file to generate a major2 object for rendering the sidebar

packages/frontend-v2/pages/home.tsx
```diff
@@ -65,13 +77,14 @@ const HomePage: NextPage = () => {
...
-    // render just the boiler plate page since we couldn't fetch the student w plans
-    return <PageLayout />;
+    // If we couldn't fetch the student's plan, show a blank page for now.
+    // We might want to show some more actionable error in the future.
+    return <div></div>;
   }
 
   // handle loading state
   if (!student) {
-    return <LoadingPage pageLayout={PageLayout} />;
```
AlpacaFur, line 74: Previously we loaded the entire page as a skeleton UI of sorts, this had the unintended side-effect of causing the Sidebar to mount multiple times, resulting in us sending twice as many queries to Search's API. Without the skeleton UI, the page shows a loading spinner for the <1s that the page takes to load in.

packages/frontend-v2/components/Sidebar/Sidebar.tsx
```diff
@@ -0,0 +1,292 @@
...
+            dndIdPrefix={dndIdPrefix + "-" + index}
+            key={index}
+          />
+        ))}
+      </Box>
+    </Box>
+  );
+};
+
+const SectionRequirement: React.FC<SidebarRequirementProps> = ({
```
BrandonLim8890, line 176: I think this + all the cases below could be its own separate file since the logic for it is kind of grouped together, but up to you to decide!

packages/frontend-v2/components/Sidebar/Sidebar.tsx
```diff
@@ -0,0 +1,292 @@
...
+      case "OR":
+        return renderOR(requirement);
+      case "RANGE":
+        return renderRange(requirement);
+      case "COURSE":
+        return renderCourse(requirement);
+      case "SECTION":
+        return renderSection(requirement);
+      default:
+        return <p>Unsupported type</p>;
```
BrandonLim8890, line 196: I think throwing an error here is better than returning unsupported type. Should be some sort of assertUnreachable

packages/frontend-v2/pages/home.tsx
```diff
@@ -65,13 +77,14 @@ const HomePage: NextPage = () => {
     logger.error("HomePage", error);
     handleApiClientError(error, router);
 
-    // render just the boiler plate page since we couldn't fetch the student w plans
-    return <PageLayout />;
+    // If we couldn't fetch the student's plan, show a blank page for now.
+    // We might want to show some more actionable error in the future.
+    return <div></div>;
```
BrandonLim8890, line 82: Is there a reason why you guys got rid of the `<PageLayout />`?

BrandonLim8890 (commented):
Looks good overall!
I think it should be good but I'm a little unclear on the purpose of the PR. Is it to fully implement the sidebar with partial API calls or just to get the sidebar portion working?

Couple of small things I noticed: 
1. I think the UI/UX is pretty weird where you drag a course from the sidebar into the plan, and the course stays in the plan but also moves back into the sidebar. Probably going to be another ticket for us to fix after Alpha
2. I feel like the sidebar should not be available until all courses are loaded. From a UX perspective I think this makes way more sense rather than trying to drag courses in whilst the sidebar is still moving/populating.

Overall, fix #2 + comments and it looks good to merge. And update the PR description with more detailed scope of what this PR is going to do, and then what the subsequent PR will be fixing.

packages/frontend-v2/components/Sidebar/Sidebar.tsx
```diff
@@ -0,0 +1,292 @@
...
+        getRequiredCourses(section.requirements, requiredCourses);
+        return courses.concat(requiredCourses);
+      },
+      []
+    );
+    // console.info("Making", requirements.length, "requests")
+
+    const promises: Promise<ScheduleCourse2<null> | null>[] = [];
+
+    for (const requirement of requirements) {
```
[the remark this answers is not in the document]
jia614: i might be misunderstanding your comment, but i believe this loop is being executed in the useEffect. did you mean something else?

packages/frontend-v2/pages/home.tsx
jia614: having page layout here would render the sidebar rather than an empty page. i don't think either is necessarily incorrect, just not sure yet how we want to display to a user that their plan isn't loaded. i don't think it makes a huge difference either way at the moment tho

jia614:
@BrandonLim8890 tanks for da comments :)

i think #2 will be addressed by one of our next PRs (to integrate search's new batch endpoint) since we won't be requesting every course one by one anymore. i've also updated this pr's desc so lmk if there's anything that's unclear or missing

packages/frontend-v2/components/Sidebar/Sidebar.tsx
```diff
@@ -0,0 +1,292 @@
...
+        //   });
+        // })
+        SearchAPI.fetchCourse(
+          requirement.subject,
+          requirement.classId.toString()
+        )
+      );
+    }
+
+    Promise.all(promises).then((courses) => {
```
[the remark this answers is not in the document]
AlpacaFur: I believe this is already in the useEffect. We considered using async await, but it would involve making another function within the useEffect that we immediately call, because the top-level useEffect callback isn't allowed to be async. Since we don't have any nested promises in this case, we think leaving it as .then is probably okay in this case.

packages/frontend-v2/components/Sidebar/Sidebar.tsx
AlpacaFur: Good idea, we refactored the Sidebar file into SidebarSection and SectionRequirement.

packages/frontend-v2/components/Sidebar/Sidebar.tsx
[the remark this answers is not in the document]
BrandonLim8890: IM BLIND

Nickdaco (commented):
Sidebar lookin gooooood!

I think the having the course jump back to the sidebar after placing is a little confusing (but I'm not sure if that's being handled in a different pr)

Other than that the code and everything else looks pretty good!

BrandonLim8890 (approved):
LGTM!

PR #491, opened 2022-10-30
Course Drag-and-Drop Performance and QoL Improvements

### LINKED ISSUE #471: Improve DND Performance
# Context
- We're having performance issues with dnd when we render the sidebar.
- Specifically, dragging is slow.
- We found out that this is because of the amount of html content within a component.

# Possible solutions
- Hide trash button + content that can be hidden in a draggable course. Only show it when necessary: on hover, like the previous app. THIS IS PROVEN TO HELP A LOT.

# Other solutions if we still need improvement
- Virtualized lists to prevent all draggable courses to be a part of the dom
- Only allow one requirement section to be expanded at a time so that a massive requirement section doesn't bottleneck the performance

### LINKED ISSUE #485: Fix Sidebar Course Dragging
For some reason when trying to drag a course fast out the sidebar, the course doesn't actually drag?

AlpacaFur:
# Description

A bunch of miscellaneous performance and QoL improvements for drag-and-drop. (Closes #471)

- Converted the ScheduleCourse component from Chakra components to (mostly) standard HTML in order to significantly improve drag-and-drop performance.
- Prevent collapsed course categories in the sidebar from being put into the DOM at all. Previously they were set to `display: none` but were still in the DOM, which slowed down drag-and-drop.
- Fix drop animation where previously a course dragged from the sidebar would fly back into the sidebar as well as being put into the schedule. Now the course just goes into the schedule.
- Adjusted the course drag-and-drop collision algorithm to be more intuitive. It now chooses the droppable that the mouse is over and falls back to rectangle collision if the mouse isn't over anything. Previously it was just rectangle collision which felt a little harder to use/predict.
- Added a minor transition effect to course hover and semester drag-over.
- Fixed the draggable hitbox for courses so can no longer miss the draggable part of the element. Previously if you tried to drag the top part of the element, it wouldn't actually drag. (Closes #485)

https://user-images.githubusercontent.com/2746893/198897617-07c0df0e-fac9-4daf-95f5-7da28baf9695.mp4

## Type of change

# How Has This Been Tested?

This has been manually tested by expanding all categories and adding 6 courses to each semester in order to test worst-case performance.

# Checklist:


packages/frontend-v2/components/ScheduleCourse/ScheduleCourse.tsx
```diff
@@ -57,6 +55,7 @@ interface ScheduleCourseProps extends DraggableScheduleCourseProps {
   attributes?: any;
   transform?: string;
   isDisabled: boolean;
+  isOverlay?: boolean;
```
AlpacaFur, line 58: I had to add a type to the props here because for whatever reason isDragging from the useDraggable hook is briefly the opposite of what it should be when the drag starts, causing some visual jank.

packages/frontend-v2/components/ScheduleCourse/ScheduleCourse.tsx
```diff
@@ -69,52 +68,116 @@ export const ScheduleCourse = forwardRef<
...
+              flexShrink: 0,
+              color: "var(--chakra-colors-primary-blue-dark-main)",
+              verticalAlign: "middle",
+            }}
+          >
+            <path
+              fill="currentColor"
+              d="M3,2 C2.44771525,2 2,1.55228475 2,1 C2,0.44771525 2.44771525,0 3,0 C3.55228475,0 4,0.44771525 4,1 C4,1.55228475 3.55228475,2 3,2 Z M3,6 C2.44771525,6 2,5.55228475 2,5 C2,4.44771525 2.44771525,4 3,4 C3.55228475,4 4,4.44771525 4,5 C4,5.55228475 3.55228475,6 3,6 Z M3,10 C2.44771525,10 2,9.55228475 2,9 C2,8.44771525 2.44771525,8 3,8 C3.55228475,8 4,8.44771525 4,9 C4,9.55228475 3.55228475,10 3,10 Z M7,2 C6.44771525,2 6,1.55228475 6,1 C6,0.44771525 6.44771525,0 7,0 C7.55228475,0 8,0.44771525 8,1 C8,1.55228475 7.55228475,2 7,2 Z M7,6 C6.44771525,6 6,5.55228475 6,5 C6,4.44771525 6.44771525,4 7,4 C7.55228475,4 8,4.44771525 8,5 C8,5.55228475 7.55228475,6 7,6 Z M7,10 C6.44771525,10 6,9.55228475 6,9 C6,8.44771525 6.44771525,8 7,8 C7.55228475,8 8,8.44771525 8,9 C8,9.55228475 7.55228475,10 7,10 Z"
+            ></path>
+          </svg>
```
aryanshah701, line 133: could we move this into its own file/component just for readability

packages/frontend-v2/pages/home.tsx
```diff
@@ -35,6 +38,18 @@ import { getMajor2Example } from "../utils/convertMajor";
...
+// See https://docs.dndkit.com/api-documentation/context-provider/collision-detection-algorithms for more info.
+const courseDndCollisisonAlgorithm: CollisionDetection = (args) => {
+  const pointerCollisions = pointerWithin(args);
+  if (pointerCollisions.length > 0) {
+    return pointerCollisions;
+  } else {
+    // Fallback as recommended by the dnd-kit docs
+    return rectIntersection(args);
+  }
+};
```
aryanshah701, line 51: oooh naiceee

packages/frontend-v2/utils/plan/updatePlanOnDragEnd.ts
```diff
@@ -52,7 +52,8 @@ export const updatePlanOnDragEnd = (
       }
 
       if (oldTerm === newTerm) {
-        throw new Error("Course is being dragged over its own term");
+        // Course is being dragged over its own term, so we do nothing
+        return;
```
aryanshah701, line 56: any reason we removed this error? the way updatePlanOnDragEnd is used in home is that we continue with the update if the function doesn't throw an error, don't see why we need to continue with the update if the course is being dragged over its own term

packages/frontend-v2/components/ScheduleCourse/ScheduleCourse.tsx
aryanshah701: lmfao what even

aryanshah701 (approved):
SUCH A CRACKED PR!!! Really awesome work here, lowkey think devs could benefit from hearing about your journey into debugging and fixing this issue... amazing stuff, it works like a charm.

Love the PR description too, a really nice example of a concise but precise description. Reduced the review time by a drastic amount!

Just 2 minor comments, ship it after! ⛴

packages/frontend-v2/components/ScheduleCourse/ScheduleCourse.tsx
AlpacaFur: Good call, will do.

packages/frontend-v2/utils/plan/updatePlanOnDragEnd.ts
AlpacaFur: Ah I was confused why we were throwing for a non-fatal error that we handled, but that if it's used in control flow above this function, that makes sense. I'll revert that change and leave a comment in the function description.

PR #493, opened 2022-10-30
Prevent course duplicates in a semester

### LINKED ISSUE #478: Fix Deleting Duplicate Courses In Plan
All duplicate courses get deleted if one course is delete. Fix this to only delete the one being deleted.

Edit: We should simply prevent duplicate courses in 1 semester lmao.

aryanshah701:
# Description

https://user-images.githubusercontent.com/30478978/198899604-8d52e8c3-2539-41cd-a0c8-05000cf4664e.mov 

- Currently, we allow duplicate courses to be dragged in a semester. That's weird, this prevents that.
- We also currently don't allow duplicate courses in a plan with the add course button. We can allow this and instead only prevent duplicate courses in a semester.

Closes #478 

## Type of change

# How Has This Been Tested?

Manually. See video.

# Checklist:


packages/frontend-v2/components/AddCourseModal/AddCourseModal.tsx
```diff
@@ -28,8 +28,8 @@ interface AddCourseModalProps {
   /** Function to close the modal UX, returned from the useDisclosure chakra hook */
   closeModalDisplay: () => void;
 
-  /** Function to check if the given cousrs exists in the plan being displayed. */
-  isCourseInCurrPlan: (course: ScheduleCourse2<unknown>) => boolean;
```
aryanshah701, line 32: name change, that'll propagate a few components

packages/frontend-v2/components/Plan/ScheduleTerm.tsx
```diff
@@ -103,27 +103,6 @@ const ScheduleTermHeader: React.FC<ScheduleTermHeaderProps> = ({
...
-
-  const seasonDisplayWord = SEASON_TO_SEASON_DISPLAY_WORD.get(season);
-  if (!seasonDisplayWord) {
-    logger.debug("getSeasonDisplayWord", "Unknown season", season);
-    throw new Error("Unknown Season");
-  }
-
-  return seasonDisplayWord;
-};
-
```
aryanshah701, line 126: moved to utils for reusability

packages/frontend-v2/utils/course/isEqualCourses.ts
```diff
@@ -1,8 +1,14 @@
-import { ScheduleCourse2 } from "@graduate/common";
+/** Only the details we need to compare courses. */
+type CourseEqualsDetails = {
+  classId: string;
+  subject: string;
+} & {
+  [key: string]: any;
+};
```
aryanshah701, line 7: kinda weird, but essentially allows us to pass in any object that has classId and subject. it's nice so that we can use this when we just have classId and subject annnnd we don't have to extract classId and subject from ScheduleCourse if we have the actual schedule course

AlpacaFur (commented):
Currently there's an edge case where adding duplicate co-reqs is not prevented. For example, add CS1800 to a semester, which adds CS1800 and CS1802. Then remove only CS1800 and re-add it using the same add course modal. (Results in two CS1802s). You could get around this by filtering the co-reqs by whether they're already in the term before you add them in the modal.

packages/frontend-v2/utils/plan/updatePlanOnDragEnd.ts
```diff
@@ -38,23 +47,52 @@ export const updatePlanOnDragEnd = (
...
+      !isSameTerm
+    ) {
+      toast.error(
+        `Oops, ${getCourseDisplayString(
+          draggedCourseDetails
+        )} already exists in Year ${newTerm.year}, ${getSeasonDisplayWord(
+          newTerm.season
+        )}.`
       );
+      throw new Error("Duplicate course in term.");
```
AlpacaFur, line 79: As a general question, should we be throwing errors for user-error that we've successfully handled?

packages/frontend-v2/utils/plan/updatePlanOnDragEnd.ts
```diff
@@ -38,23 +47,52 @@ export const updatePlanOnDragEnd = (
...
+    }
 
+    /*
+     * Course is from a term, so we need to move it, we don't need to move
+     * courses that are from the sidebar.
+     */
+    if (!draggedCourse.data.current.isFromSidebar) {
```
AlpacaFur, line 86: This could reuse the isFromSidebar constant from above.

packages/frontend-v2/utils/course/isEqualCourses.ts
```diff
@@ -1,8 +1,14 @@
...
```
AlpacaFur, line 8: Just as a clarification, is there a benefit to specifying string keys for extra properties versus just doing `& any`?

packages/frontend-v2/utils/course/getCourseDisplayString.ts
```diff
@@ -1,9 +1,15 @@
-import { ScheduleCourse2 } from "@graduate/common";
+/** Only the details we need to display a course. */
+type CourseDisplayDetails = {
```
AlpacaFur, line 2: This type is identical to CourseEqualDetails, so we might want to make it just one type in our types file that we use for both (and for potential future uses as well).

packages/frontend-v2/components/Plan/ScheduleTerm.tsx
```diff
@@ -32,14 +29,17 @@ interface ScheduleTermProps {
...
   scheduleTerm,
-  isCourseInCurrPlan,
   addClassesToTermInCurrPlan,
   removeCourseFromTermInCurrPlan,
   isLastColumn,
 }) => {
   const { isOver, setNodeRef } = useDroppable({ id: scheduleTerm.id });
   const { isOpen, onOpen, onClose } = useDisclosure();
 
+  const isCourseInCurrTerm = (course: ScheduleCourse2<unknown>) => {
```
AlpacaFur, line 39: As a clarification, is unknown being used as the generic here because we don't care about the type of the `id` property?

AlpacaFur (commented):
I made one comment about an edge case where it's still possible to have a duplicate course, but other than that, looks good! I also left a few questions for some stuff that I was wondering about.

packages/frontend-v2/utils/plan/updatePlanOnDragEnd.ts
aryanshah701: Currently it's just the way the updatePlanOnDragEnd API works where we throw errors to indicate "do not update the plan and make the API post request".

In most cases we don't even need to alert the user of the error though(dragging over a term not found etc.). So in those cases we just throw the error. In this case, we want to alert the user so we alert and throw.

A potential solution around this is as follows:
- Don't throw errors, and instead return them
- Return a flag indicating whether we want to alert the user of the error and then let the calling function take care of it

Not super sure if this is a huge improvement, we can discuss this and add this as a tech debt issue if there's benefit. But thanks for pointing it out!!

packages/frontend-v2/utils/plan/updatePlanOnDragEnd.ts
aryanshah701: my badd, thank you!!

packages/frontend-v2/utils/course/isEqualCourses.ts
aryanshah701: Yeap! With & any the following would be fine by TS
```
isEqualCourses({ yo: "hi" }, { yo: 2 });
```

Because the & any essentially makes the whole type any. But the way it currently is makes it so that at least classId and subject are required. But other keys could also exist.

packages/frontend-v2/utils/course/getCourseDisplayString.ts
aryanshah701: Considered it but that couples the implementation of both functions. The types happen to be the same because we compare courses by the same stuff we use to generate their string. This could very well change in the future

packages/frontend-v2/components/Plan/ScheduleTerm.tsx
aryanshah701: precisely! this allows us to call the function with or without a type. and it also prevent the function itself from accessing the id type(without explicitly checking its type) because it doesn't know what it is.

packages/frontend-v2/components/AddCourseModal/AddCourseModal.tsx
```diff
@@ -74,8 +74,15 @@ export const AddCourseModal: React.FC<AddCourseModalProps> = ({
...
 
-    // push course + any coreqs of the course
-    const coreqs = await getRequiredCourseCoreqs(course);
+    // grab any coreqs of the course that haven't already been selected/added to the term
+    const coreqs = (await getRequiredCourseCoreqs(course)).filter((coreq) => {
+      const isAlreadySelected = selectedCourses.find((selectedCourse) =>
+        isEqualCourses(selectedCourse, coreq)
+      );
+      const isAlreadyAdded = isCourseInCurrTerm(coreq);
+      return isAlreadyAdded || isAlreadySelected;
```
AlpacaFur, line 83: I think this condition is reversed at the moment (only keeps courses that are already added or already selected). I tested it with the suggested change and it works as intended.

```suggestion
      return !(isAlreadyAdded || isAlreadySelected);
```

AlpacaFur (changes requested):
I made a suggestion about a logic tweak but once that's done, looks good!

packages/frontend-v2/components/AddCourseModal/AddCourseModal.tsx
aryanshah701: Thank you so so much, I seriously need to stop doing graduate dev at work, my brain is making some very dumb mistakes

AlpacaFur (approved):
🚢🚢🚢

PR #495, opened 2022-10-31
Add credit information to sidebar class data

### LINKED ISSUE #492: Semester Credits Display as NaN in the Plan
Screen Shot 2022-10-30 at 3 38 38 PM

Semester credits display as NaN in the plan when adding a course from the sidebar (or having preexisting courses from the sidebar). Currently reproducable on main.

AlpacaFur:
# Description

Small fix to get class credits information from Search's API. Also made missing courses display their subject + classId so it's easier to debug missing classes later and provides some extra info to users.

Closes #492

## Type of change

# How Has This Been Tested?
Manually tested by checking that the credits add up properly in the semesters now.

# Checklist:

### DIFF
--- a/packages/api-client/index.ts
+++ b/packages/api-client/index.ts
@@ -87,6 +87,21 @@ interface SearchClass {
   minCredits: number;
 }
 
+function occurrenceToCourse(occurrence): ScheduleCourse2<null> | null {
+  if (!occurrence) return null;
+
+  return {
+    name: occurrence.name,
+    subject: occurrence.subject,
+    classId: occurrence.classId,
+    numCreditsMax: occurrence.maxCredits,
+    numCreditsMin: occurrence.minCredits,
+    prereqs: occurrence.prereqs,
+    coreqs: occurrence.coreqs,
+    id: null,
+  };
+}
+
 /**
  * A client for interacting with the Search API. Allows us to fetch and search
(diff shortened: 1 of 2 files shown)


packages/api-client/index.ts
```diff
@@ -163,7 +165,17 @@ class SearchAPIClient {
...
     if (coursesData.bulkClasses) {
-      return coursesData.bulkClasses;
+      return coursesData.bulkClasses.map((course) => {
+        if (course === null) return null;
+
+        return {
+          name: course.name,
+          classId: course.classId,
+          subject: course.subject,
+          numCreditsMin: course?.latestOccurrence?.minCredits,
```
AlpacaFur, line 175: I had to unwrap the data here bc minCredits is nested within latestOccurence (and we have a different name for it).

aryanshah701: awesome, we do similar logic in the fetchCourse function... could you abstract it out? We use fetchCourse in getting course coreqs and things I think(may be time to move that to bulk as well but that's for a later task)

aryanshah701 (approved):
🚢🚢🚢

packages/api-client/index.ts
AlpacaFur: will do!

PR #501, opened 2022-11-06
Frontend Coreq and Prereq Errors

### LINKED ISSUE #454: Display Co-req And Pre-req Errors In Plan

Nickdaco:
# Description

Added coreq/prereq errors for classes. Displays error icon in the year as well as on the specific class. The error Icon on that class opens a modal that tells you specifically what classes you can take to satisfy the error.

Next steps are to add more styling and pagination to the modal.
Another step is also to address the Prop drill down issue, where we would instead use context for the errors.

https://user-images.githubusercontent.com/20482676/200195253-07339502-a79b-414f-8621-19a9c4328bfe.mp4

Closes #454

## Type of change

# How Has This Been Tested?

Tested with dragging in coreq/prereq errors and checking if they were giving the correct errors as well as if they went away when we added the correct classes.

# Checklist:

### DIFF
--- a/packages/frontend-v2/components/Plan/ReqErrorModal.tsx
+++ b/packages/frontend-v2/components/Plan/ReqErrorModal.tsx
@@ -31,13 +31,26 @@ export const ReqErrorModal: React.FC<ReqErrorModalProps> = ({
 }) => {
   const { isOpen, onOpen, onClose } = useDisclosure();
   return (
-    <>
+    <Flex
+      justifySelf="stretch"
+      alignSelf="stretch"
+      _hover={{
+        background: "primary.red.main",
+        fill: "white",
+        svg: { color: "white" },
+      }}
+      _active={{ background: "primary.red.900" }}
+      onClick={() => {
+        onOpen();
+      }}
+    >
       <WarningIcon
         color="primary.red.main"
-        transition="color 0.1s ease"
-        onClick={() => {
-          onOpen();
-        }}
+        width="2rem"
+        alignSelf="center"
+        alignItems="center"
+        justifySelf="center"
+        transition="background 0.15s ease"
       />
       <Modal isOpen={isOpen} onClose={onClose}>
         <ModalOverlay />
@@ -77,7 +90,7 @@ export const ReqErrorModal: React.FC<ReqErrorModalProps> = ({
           </ModalFooter>
         </ModalContent>
       </Modal>
-    </>
+    </Flex>
   );
 };
 
--- a/packages/frontend-v2/components/Plan/ReqErrorModal.tsx
+++ b/packages/frontend-v2/components/Plan/ReqErrorModal.tsx
@@ -1,0 +1,79 @@
+import { DeleteIcon } from "@chakra-ui/icons";
+import {
+  Button,
+  IconButton,
+  Modal,
+  ModalBody,
+  ModalCloseButton,
+  ModalContent,
+  ModalFooter,
+  ModalHeader,
+  ModalOverlay,
+  useDisclosure,
+} from "@chakra-ui/react";
+import { API } from "@graduate/api-client";
+import { useRouter } from "next/router";
+import { mutate } from "swr";
+import { USE_STUDENT_WITH_PLANS_SWR_KEY } from "../../hooks";
+import { handleApiClientError } from "../../utils";
+import { GrayButton } from "../Button";
+
+interface DeletePlanModalProps {
+  planName: string;
+  planId: number;
+  onClose: (isDeleted: boolean, closeDisplay: () => void) => void;
+}
+export const DeletePlanModal: React.FC<DeletePlanModalProps> = ({
+  planName,
+  planId,
+  onClose,
+}) => {
+  const router = useRouter();
+  const { onOpen, onClose: closeDisplay, isOpen } = useDisclosure();
+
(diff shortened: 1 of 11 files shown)


packages/frontend-v2/components/Plan/ReqErrorModal.tsx
```diff
@@ -0,0 +1,136 @@
...
+    </Flex>
+  );
+};
+
+interface ParseCourseProps {
+  course: INEUPrereqError | undefined;
+}
+
+// Look through the course error until there are no more errors!
+const ParseCourse: React.FC<ParseCourseProps> = ({ course }) => {
```
Nickdaco, line 102: React component is recursive because the error type is recursive

packages/frontend-v2/components/Plan/ReqErrorModal.tsx
```diff
@@ -0,0 +1,165 @@
...
+          </ModalBody>
+
+          <ModalFooter>
+            <Flex
+              justifyContent="space-between"
+              width="100%"
+              alignItems="center"
+            >
+              <Button
+                variant="solidRed"
```
AlpacaFur, line 97: I think from a UI perspective, we should probably keep red buttons for dangerous actions (closing this modal doesn't have any destructive effects).
```suggestion
                variant="solidBlue"
```

packages/frontend-v2/components/Plan/ReqErrorModal.tsx
```diff
@@ -0,0 +1,165 @@
...
+              {coReqErr && preReqErr && (
+                <Text fontSize="sm">{`Page ${page + 1} of 2`}</Text>
+              )}
+              {coReqErr && preReqErr && (
+                <Button
+                  variant="solidBlue"
+                  mr={3}
+                  onClick={() => setPage(page == 0 ? 1 : 0)}
+                >
+                  {page == 0 ? "Next" : "Prev"}
```
AlpacaFur, line 115: This is something I'll probably ask about in more detail during GM, but I'm not sure that paginating the errors is necessary here because we will only ever have two (Coreq and Prereq). Right now it seems like it just adds an extra click to see the other error, but I'm not sure of the initial motivation behind it. If there are some really long errors that would make stacking both errors infeasible we might want to consider another solution (maybe make the modal scrollable?).

packages/frontend-v2/utils/plan/preAndCoReqCheck.ts
```diff
@@ -1,96 +1,135 @@
 import {
-  courseError, INEUAndPrereq, INEUOrPrereq, INEUPrereq, INEUPrereqCourse, INEUPrereqError, preReqWarnings, Schedule2, ScheduleTerm2, courseToString
+  CoReqWarnings,
+  courseToString,
+  INEUAndPrereq,
+  INEUOrPrereq,
+  INEUPrereq,
+  INEUPrereqCourse,
+  INEUPrereqError,
```
AlpacaFur, line 8: Since we're using them for both prereqs and coreqs, we should probably replace the word prereq with just req in the INEU*Prereq types for the sake of clarity. (I know these type names predate this PR but now is probably a good time to change them).

AlpacaFur (changes requested):
I made a few small suggestions but other than those, looks good!

packages/frontend-v2/utils/plan/preAndCoReqCheck.ts
BrandonLim8890: True okay will amend

BrandonLim8890: True okay will amend

AlpacaFur (approved):
Looks good!

packages/frontend-v2/components/ScheduleCourse/ScheduleCourse.tsx
```diff
@@ -1,13 +1,20 @@
...
+  ScheduleCourse2,
+} from "@graduate/common";
+import { forwardRef, useState } from "react";
+import { ReqErrorModal } from "../Plan/ReqErrorModal";
 import { CourseDragIcon } from "./CourseDragIcon";
 
 interface DraggableScheduleCourseProps {
   scheduleCourse: ScheduleCourse2<string>;
+  coReqErr: INEUReqError | undefined;
+  preReqErr: INEUReqError | undefined;
```
AlpacaFur, line 17:
```suggestion
  coReqErr?: INEUReqError | undefined;
  preReqErr?: INEUReqError | undefined;
```
Lint fix part 1

packages/frontend-v2/components/ScheduleCourse/ScheduleCourse.tsx
```diff
@@ -19,6 +26,8 @@ interface DraggableScheduleCourseProps {
 export const DraggableScheduleCourse: React.FC<
   DraggableScheduleCourseProps
 > = ({
+  coReqErr,
+  preReqErr,
```
AlpacaFur, line 30:
```suggestion
  coReqErr = undefined,
  preReqErr = undefined,
```
Lint fix part 2. Basically a draggable course doesn't always have coReq/preReq errors and one of it's usages in the sidebar was erroring in the lint.

AlpacaFur (approved):
Small tweak to fix the lint error.

PR #506, opened 2022-11-29
Fix Major Validation Double-Counting

### LINKED ISSUE #503: Validation, Fix: Prevent Course Double-Counting
Right now, it is somehow possible for a course to simultaneously fulfill a requirement in a section and a course in a xom (for example, CS Capstone and CS Elective in the demo BSCS plan). It seems to be conditional on whether the xom is in the last unfulfilled section. Here is an example of the bug:

Plan:
```
section("Capstone", 1, [
  or(course("CS", 4100), course("CS", 4300))
]),
section("Elective", 1, [
  xom(8, [range(0, "CS", 2500, 5010, [])])
]),
section("Presentation", 1, [
  course("THTR", 1170)
])
```

Courses:
CS4100 and CS4300.

Currently, there is only an error on the Presentation section. Here we would expect there to be an error on the Elective section as well, because otherwise one of the courses is being double-counted in both the Capstone OR and the Elective XOM. 

Interestingly, if we remove the Presentation section, we get an error (AND_NO_SOLUTION) on the Elective section as expected.

AlpacaFur:
# Description

https://user-images.githubusercontent.com/2746893/204439039-02577a94-3157-4445-acc9-061a3dfd3c01.mp4

Fixes an edge case where courses could be double counted because major validation could only return either an AndUnsatRequirements error or a AndNoSolution error but not both. As a result if there were any unsatisfied children in an And it would mask the NoSolution error, resulting in not enough errors being returned. This caused the UI to check off both sections despite such a solution being impossible.

Also makes some code clarity changes to how the section errors are propagated.

Closes #503 

## Type of change

# How Has This Been Tested?

This has been manually tested and some automated tests were added in `major2-validation.test.ts` that cover the case that this fixes.

# Checklist:

### DIFF
--- a/packages/common/src/major2-validation.ts
+++ b/packages/common/src/major2-validation.ts
@@ -55,17 +55,18 @@ export const MajorValidationErrorType = {
     Type: "AND",
     UnsatChild: "AND_UNSAT_CHILD",
     NoSolution: "AND_NO_SOLUTION",
+    UnsatChildAndNoSolution: "AND_UNSAT_CHILD_AND_NO_SOLUTION",
   },
   Or: "OR",
   XofMany: "XOM",
   Section: "SECTION",
 } as const;
-type ChildError = MajorValidationError & { childIndex: number };
+export type ChildError = MajorValidationError & { childIndex: number };
 type CourseError = {
   type: typeof MajorValidationErrorType.Course;
   requiredCourse: string;
 };
-const CourseError = (c: IRequiredCourse): CourseError => ({
+export const CourseError = (c: IRequiredCourse): CourseError => ({
   type: MajorValidationErrorType.Course,
   requiredCourse: courseToString(c),
 });
@@ -79,9 +80,43 @@ type AndError = {
     | {
         type: typeof MajorValidationErrorType.And.NoSolution;
         discoveredAtChild: number;
+      }
+    | {
+        type: typeof MajorValidationErrorType.And.UnsatChildAndNoSolution;
+        noSolution: {
+          type: typeof MajorValidationErrorType.And.NoSolution;
+          discoveredAtChild: number;
+        };
+        unsatChildErrors: {
+          type: typeof MajorValidationErrorType.And.UnsatChild;
+          childErrors: Array<ChildError>;
+        };
       };
 };
-const AndErrorUnsatChild = (childErrors: Array<ChildError>): AndError => ({
+
+export const AndErrorUnsatChildAndNoSolution = (
+  unsatChildErrors: Array<ChildError>,
+  noSolutionIndex: number
+): AndError => {
+  return {
+    type: MajorValidationErrorType.And.Type,
+    error: {
+      type: MajorValidationErrorType.And.UnsatChildAndNoSolution,
+      noSolution: {
+        type: MajorValidationErrorType.And.NoSolution,
+        discoveredAtChild: noSolutionIndex,
+      },
+      unsatChildErrors: {
+        type: MajorValidationErrorType.And.UnsatChild,
+        childErrors: unsatChildErrors,
+      },
+    },
+  };
+};
+
+export const AndErrorUnsatChild = (
+  childErrors: Array<ChildError>
+): AndError => ({
   type: MajorValidationErrorType.And.Type,
   error: { type: MajorValidationErrorType.And.UnsatChild, childErrors },
 });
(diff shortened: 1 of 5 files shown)


packages/common/src/major2-validation.ts
```diff
@@ -79,9 +80,43 @@ type AndError = {
...
+    | {
+        type: typeof MajorValidationErrorType.And.UnsatChildAndNoSolution;
+        noSolution: {
+          type: typeof MajorValidationErrorType.And.NoSolution;
+          discoveredAtChild: number;
+        };
+        unsatChildErrors: {
+          type: typeof MajorValidationErrorType.And.UnsatChild;
+          childErrors: Array<ChildError>;
+        };
```
AlpacaFur, line 93: This is kind of a gross type, but matches the other types and is something we can take another look at after alpha.

packages/common/src/major2-validation.ts
```diff
@@ -479,24 +514,16 @@ function validateAndRequirement(
...
 ): Result<Array<Solution>, MajorValidationError> {
-  const splitResults = validateAndSplit(r.courses, tracker);
-  const [allChildReqSolutions, childErrors] = splitResults;
-
-  // AND's children has errors
-  if (childErrors.length > 0) {
-    return Err(AndErrorUnsatChild(childErrors));
-  }
+  const results = validateRequirements(r.courses, tracker);
+  const [allChildReqSolutions, childErrors] = splitChildResults(results);
```
AlpacaFur, line 518: Validate and split separately so we can keep track of the original indices of sections for NoSolution index accuracy.

packages/common/src/major2-validation.ts
```diff
@@ -511,10 +538,26 @@ function validateAndRequirement(
...
-      return Err(AndErrorNoSolution(childIdx));
+      const actualIndex = results.findIndex((solution) => {
+        return (
+          solution.type === ResultType.Ok &&
+          solution.ok === childRequirementSolutions
+        );
+      });
+      if (childErrors.length > 0) {
+        return Err(AndErrorUnsatChildAndNoSolution(childErrors, actualIndex));
+      } else {
```
AlpacaFur, line 549: Special check for the new combined AndErrorUnsatChildAndNoSolution error (very verbose but also all important information in the name :/).

packages/common/test/major2-validation.test.ts
```diff
@@ -581,3 +596,132 @@ describe("integration suite", () => {
...
+    name: name,
+    totalCreditsRequired: totalCreditsRequired,
+    yearVersion: yearVersion,
+    requirementSections: requirementSections,
+    concentrations: {
+      minOptions: 0,
+      concentrationOptions: [],
+    },
+  };
+};
```
AlpacaFur, line 629: New helpers for testing major errors.

packages/frontend-v2/components/Sidebar/Sidebar.tsx
```diff
@@ -101,11 +101,14 @@ const Sidebar: React.FC<SidebarProps> = memo(({ major, selectedPlan }) => {
...
 
+          const sectionIsValid =
+            selectedPlan !== undefined && sectionValidationError === undefined;
+
           return (
             <SidebarSection
               key={section.title}
               section={section}
-              validationStatus={sectionValidationError}
+              isValid={sectionIsValid}
```
AlpacaFur, line 111: Since we only use the validation error for the check mark, here it's simplified to a boolean.

packages/common/src/major2-validation.ts
```diff
@@ -55,17 +55,18 @@ export const MajorValidationErrorType = {
     Type: "AND",
     UnsatChild: "AND_UNSAT_CHILD",
     NoSolution: "AND_NO_SOLUTION",
+    UnsatChildAndNoSolution: "AND_UNSAT_CHILD_AND_NO_SOLUTION",
```
BrandonLim8890, line 58: THIS NAME THO 🥲

packages/common/src/major2-validation.ts
BrandonLim8890: a little unsure of why we need to do this, but will take a look at it again to better understand

BrandonLim8890 (approved):
LGTM!

packages/common/src/major2-validation.ts
AlpacaFur: Basically we need to keep track of the index of each solution/error so that we can return the correct index for a no solution (or and unsat and no solution) error. We rely on the index in the UI to uncheck the right section. `splitChildResults` by itself returns a list of solutions and a list of errors, which are individually ordered, but we lose their position in the overall order, so we now keep track of the original order instead of just validating children and splitting in one step.

PR #507, opened 2022-11-30
Majors - Backend

aryanshah701:
# Description

## Context
- We need a solid way of dong the following things
  - Storing major objects
  - Retrieving major objects by major name and catalog year
  - Knowing which majors we support and for which years
  - Carefully update/create our plans ensuring we only allow plans that are storing the supported majors 

- This whole thing can easily be a design discussion and will be in the future. However, for now, we need to simply address these things in a way that isn't fully jank, works, and hopefully requires minimal changes in a scoped area in the future once we finalize the design.

- I think the following are sound design choices 
  - Our frontend should be able to dynamically update the supported majors list. It shouldn't rely on a static object. 
  - Our majors should be stored somewhere on the server. As we scale to 100s of majors, bundling all of them in our frontend bundle is bad.

- My approach focuses on 2 things: Simplicity since we don't have time lol, and the above 2 acceptance criteria.

## My Approach
This is how we achieve things based on my approach
  - Storing major objects: JSON for simplicity, we should move to db at some point and link plans to majors using a foreign key imo.
  - Retrieving major objects by major name and catalog year:  API endpoint, since majors are stored on the server.
  - Knowing which majors we support and for which years:
    - API endpoint, server is the source of truth for the majors we support. 
    - For now, we return a static object but in the future once the majors are in a db, we could query our db for the major names etc that we support thereby truly making it dynamic.
    -  Every time we update our db with a new major, our frontend will automatically support creating plans with it without any need for redeploying the frontend BOOM.
  - Carefully update/create our plans ensuring we only allow plans that are storing the supported majors: Validation done on the server since the server knows which majors it supports.

## Changes
- New bscs 2022 major, I should have done 2021 but did 2022 by mistake. I will add 2021 later.
- Major module for get major and get supported major endpoint.
- Update plan create and update services to validate major, year pair.
- Fix plan get endpoint. It was previously throwing a 500 when we request a plan that doesn't exist. Now it throws a forbidden, since a plan that doesn't exist isn't owned by anyone lol. 

## Upcoming work
- API client and frontend changes
- More major objects

## Type of change

# How Has This Been Tested?

I tried to write e2e API tests, failed miserably. Debugging our e2e test was taking too long so I tested manually, I created a ticket to fix e2e tests.

- Get major
major-invalid-name
major-invalid-year
major-valid

- Get supported majors
supported_majors

- Create plan
create-plan-invalid-year
create-plan-invalid-name
create-plan-valid

- Update plan
update-plan-invalid-year
update-plan-invalid-name
update-plan-valid

- [Bug fix] Get plan with invalid id
get-plan-with-invalid-id

# Checklist:


BrandonLim8890 (approved):
Haven't yet taken a look at the JSON file, but from a code perspective it looks good. Will take a deeper dive into the JSON before standup

packages/api-v2/src/major/major.controller.ts
```diff
@@ -0,0 +1,33 @@
...
+  ): Major2 {
+    const major = this.majorService.findByMajorAndYear(majorName, catalogYear);
+    if (!major) {
+      throw new NotFoundException();
+    }
+
+    return major;
+  }
+
+  @Get("supportedMajors")
```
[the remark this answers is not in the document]
aryanshah701: TRUUUEEEEE

packages/api-v2/src/major/majors/2022/bscs.json
```diff
@@ -0,0 +1,502 @@
...
+            { "type": "COURSE", "classId": 2301, "subject": "BIOL" },
+            { "type": "COURSE", "classId": 2302, "subject": "BIOL" }
+          ]
+        },
+        {
+          "type": "AND",
+          "courses": [
+            { "type": "COURSE", "classId": 2301, "subject": "BIOL" },
+            { "type": "COURSE", "classId": 2302, "subject": "BIOL" }
+          ]
```
BrandonLim8890, line 207: Duplicate from above. Can remove

packages/api-v2/src/major/majors/2022/bscs.json
```diff
@@ -0,0 +1,502 @@
...
+        {
+          "type": "OR",
+          "courses": [
+            { "type": "COURSE", "classId": 4500, "subject": "CS" },
+            { "type": "COURSE", "classId": 4530, "subject": "CS" }
+          ]
+        }
+      ],
+      "title": "Computer Science Required Courses"
+    },
```
BrandonLim8890, line 67: Missing security required courses
image

BrandonLim8890 (changes requested):
Small changes to JSON but overall looks good!

PR #526, opened 2023-01-22
Fix build issues with class transformer and api

### LINKED ISSUE #517: Fix build issues to allow babel compatibility with class transformer interceptors in API V2
- Currently, our common package houses api response classes that are used by the class transformer interceptors to clean outgoing data. 
- However, we fail to build these response types correctly and get an error when we import them from @gradaute/common in api-v2.

aryanshah701:
# Description

## Context
- We have a common package with API response types
- We use babel to build the common package
- We fail to build these types in a way that is compatible with class transformer library used in the API
- Hence, instead of using the build we simply import the types directly from the file using the path which isn't the pattern used in the codebase

## Fix
- After some stackoverflowing, I realized that we need the @babel/plugin-proposal-class-properties plugin after the @babel/plugin-proposal-decorators plugin in our babel config to build properly.

Closes #517 

## Type of change

# How Has This Been Tested?

- yarn dev:v2 works
- yarn prod:v2 works: this uses production build for common

# Checklist:

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -15,7 +15,10 @@ module.exports = function (api) {
           ],
           "@babel/preset-typescript",
         ],
-        plugins: [["@babel/plugin-proposal-decorators", { version: "legacy" }]],
+        plugins: [
+          ["@babel/plugin-proposal-decorators", { version: "legacy" }],
+          "@babel/plugin-proposal-class-properties",
+        ],
         babelrcRoots: [".", "packages/*"],
       },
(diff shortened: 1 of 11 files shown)


BrandonLim8890 (approved):
LGTM!
Just missing 3 files where the old '../../../common' is still being used, namely
`auth.service.ts`
`plan.service.ts`
`student.service.ts`.

Other than that once those are changed as well then should be good to merge

PR #530, opened 2023-01-25
Fix UI Freezes During Major Validation 

### LINKED ISSUE #504: Sidebar, Performance: Validation freezes UI with large plans.
At the moment, when you have a plan that is nearing completion, the validation algorithm starts taking significantly longer time, which completely freezes the UI until it's done. When there are only 2-3 classes left, the freeze can be 3-5 seconds of completely unresponsive UI. This is obviously not great, but we already talked about a few possible directions to explore for solutions. 

Running the algorithm server-side would mitigate the UI freeze, though we would still have to wait for a response to come back. We discussed using something like an AWS Lambda so we don't have to worry about having to allocate additional compute on  the main server. 

Another possible direction is keeping it client-side, but running it on a web worker so it's not running on the UI thread anymore.

AlpacaFur:
# Description

This PR prevents UI freezes by moving the heavy-lifting of major validation to a Web Worker which runs on a different thread. This helps keep the UI fluid while still handling validation locally. I also adjusted the UI for validation to prevent it jumping around so much. Now there are circles next to each section title regardless of status. The circle shows the current status (Valid, Error, or Loading).

Note: because React Strict Mode triggers all useEffects twice during development, the validation checkmarks look a little glitchy, but they will appear normally in production.

Closes #504 

https://user-images.githubusercontent.com/2746893/214474569-cf9ac828-efa5-4eda-b0d5-0e021a7697ed.mp4

## Type of change

# How Has This Been Tested?

This has been tested manually by trying several different schedules.

# Checklist:


packages/frontend-v2/components/Sidebar/SidebarSection.tsx
```diff
@@ -39,22 +40,28 @@ const SidebarSection: React.FC<SidebarSectionProps> = ({
...
+            background: isValid === SidebarValidationStatus.Complete ? "green" : "grey",
+            color: "white",
+            padding: "3px 6px",
+            width: "26px",
+            height: "26px",
+            display: "inline-flex",
+            alignItems: "center",
+            justifyContent: "center",
+            marginRight: "6px",
+            borderRadius: "30px"
```
aryanshah701, line 54: any reason we not using chakra here and styling with props and theme values?

packages/frontend-v2/components/Sidebar/Sidebar.tsx
```diff
@@ -64,18 +70,64 @@ const Sidebar: React.FC<SidebarProps> = memo(({ selectedPlan }) => {
...
+
+  // Set up the web worker to handle major validation for us. This helps keep the
+  // UI thread free to display our app, preventing UI freezes while our schedule
+  // is being validated.
+  useEffect(()=>{
+    if (!workerRef.current) {
+      workerRef.current = new Worker(new URL("../../worker.ts", import.meta.url));
+      workerRef.current.onmessage = (message: MessageEvent<WorkerMessage>) => {
+        switch (message.data.type) {
+          case "Loaded":
```
aryanshah701, line 97: any way we can use the enum here instead of the string?

packages/frontend-v2/worker.ts
```diff
@@ -0,0 +1,36 @@
...
+    result: validateMajor2(data.major, data.taken, data.concentration)
+  }
+
+  // Only send valdation information if it was from the latest request.
+  // This helps us avoid sending outdated information that could be sent
+  // due to race conditions.
+  if (validationPosition === currentValidationPosition) {
+    postMessage(validationResult)
+  }
+})
```
aryanshah701, line 36: seems like u might need to get prettier to run on of some of these files

packages/frontend-v2/worker.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  const validationResult: ValidationResult = {
+    type:"ValidationResult",
+    result: validateMajor2(data.major, data.taken, data.concentration)
+  }
+
+  // Only send valdation information if it was from the latest request.
+  // This helps us avoid sending outdated information that could be sent
+  // due to race conditions.
+  if (validationPosition === currentValidationPosition) {
```
aryanshah701, line 33: oooh nice

aryanshah701 (approved):
Reallyyyy cool stuff!! Thank you for learning this and picking it up, not only are the results so solid but it's pretty cool tech that you just taught me!

Shippp it, just few minor nits here are and there. Can't wait to see it on staging 🤩

packages/frontend-v2/components/Sidebar/Sidebar.tsx
```diff
@@ -54,6 +60,9 @@ const getRequiredCourses = (
   }
 };
 
+// A number to help avoid displaying stale validation info.
+let currentRequestNum = 0;
```
AlpacaFur, line 64: Turns out that my previous strategy to avoid race conditions didn't work due to a quirk of web workers, so I moved the logic into the sidebar, which functions correctly. This did uncover the fact that the worker will do all previously requested validations before starting a new one, so with large plans, the user can end up waiting for a long time, unnecessarily. 

I think in a potential future ticket, I'm going to look into the performance implications of spawning a new worker each time, which would allow us to potentially cancel the useless stale validations entirely.

packages/frontend-v2/components/Sidebar/SidebarSection.tsx
```diff
@@ -38,25 +42,42 @@ const SidebarSection: React.FC<SidebarSectionProps> = ({
         _active={{
           backgroundColor: "neutral.200",
         }}
+        display="flex"
+        position="sticky"
```
AlpacaFur, line 46: Made the sidebar section headers sticky, so you can see what section you're in even when you've scrolled down :)

packages/frontend-v2/validation-worker/worker-messages.ts
```diff
@@ -0,0 +1,22 @@
+import { Major2, MajorValidationResult, ScheduleCourse2 } from "@graduate/common"
+
```
AlpacaFur, line 2: Split the types out into a separate file and put them into a better location.

packages/frontend-v2/validation-worker/worker.ts
```diff
@@ -0,0 +1,16 @@
...
+
+addEventListener("message", ({data}: MessageEvent<WorkerPostInfo>) => {
+  const validationResult: ValidationResult = {
+    type: WorkerMessageType.ValidationResult,
+    result: validateMajor2(data.major, data.taken, data.concentration),
+    requestNumber: data.requestNumber
+  }
+  
+  postMessage(validationResult)
+})
```
AlpacaFur, line 16: Without the types and race condition logic, the actual worker script is very simple now :)

packages/frontend-v2/components/Sidebar/Sidebar.tsx
AlpacaFur: I made a new issue that include some thoughts for future validation performance improvements (#535).

aryanshah701: ahh that's annoying, thanks for making the ticket and figuring this out!!

packages/frontend-v2/components/Sidebar/SidebarSection.tsx
aryanshah701: NAICE

PR #537, opened 2023-02-04
Brandon/add delete drag

### LINKED ISSUE #523: Delete course by dragging outside of the schedule
Support deleting a course by dragging it anywhere outside the schedule. Get funky with animations for brownie points.

BrandonLim8890:
# Description

This PR adds in functionality to the drag and drop flow, allowing users to remove courses when they are dragged outside the plan. This PR also includes preliminary designs for how the draggable course will look like when dragged outside of the plan, but there will be a follow up PR to update the designs when Evelyn + Angela come up with finalized designs. 

https://user-images.githubusercontent.com/20124104/216783074-2169e941-a333-412d-922f-2a4634f903c4.mov

Closes #523 

## Type of change

# How Has This Been Tested?

Dragging the course outside of the plan deletes it, but anywhere in the plan, it will get deleted.

# Checklist:


packages/frontend-v2/components/Plan/Plan.tsx
```diff
@@ -40,6 +43,8 @@ export const Plan: React.FC<PlanProps> = ({
     }
   };
 
+  const { setNodeRef } = useDroppable({ id: 'plan' });
```
BrandonLim8890, line 46: Here we make the plan a droppable. This is because our implementation of delete is performed when a schedule course is dragged over anything that isn't a droppable. Thus, we want to make it so that anytime a course is dragged over the plan, but not a term within the plan, it does not get deleted.

packages/frontend-v2/components/Plan/Plan.tsx
```diff
@@ -11,11 +11,13 @@ import { useState } from "react";
 import { addClassesToTerm, removeYearFromPlan } from "../../utils";
 import { removeCourseFromTerm } from "../../utils/";
 import { ScheduleYear } from "./ScheduleYear";
+import { useDroppable } from "@dnd-kit/core";
 
 interface PlanProps {
   plan: PlanModel<string>;
   preReqErr?: PreReqWarnings;
   coReqErr?: CoReqWarnings;
+  setIsRemove?: (val: boolean) => void
```
BrandonLim8890, line 20: This is passed all the way down to the schedule course that is being rendered. This is the differentiator between the initial schedule course, and the one that DND kit creates when you start dragging. We use this `isRemove` value and `setIsRemove` to essentially tell react that **this schedule course** is the one that is being dragged, and if it is not over a droppable, style it a specific way to indicate to the user that it is being removed.

packages/frontend-v2/pages/home.tsx
```diff
@@ -252,6 +250,7 @@ const HomePage: NextPage = () => {
             isOverlay={true}
             coReqErr={undefined}
             preReqErr={undefined}
+            isRemove={isRemove}
```
BrandonLim8890, line 253: This is where the value of `isRemove` is actually used to style the additional course we render

packages/frontend-v2/utils/plan/updatePlanOnDragEnd.ts
```diff
@@ -29,82 +29,95 @@ import { isCourseInTerm } from "./isCourseInTerm";
...
-    const newTerm = scheduleTerms.find(
-      (term) => term.id === draggedOverTerm.id
-    );
-
-    if (!newTerm) {
-      throw new Error("Term the course is dragged over isn't found");
-    }
+    // If it is not being dragged over a droppable, then we delete the course
+    if (draggedOverTerm == null) {
+      const newTerm = scheduleTerms.find(term => term.classes.some((course) => course.id === draggedCourse.id));
```
jia614, line 40: this is actually the old term right? nit but could u change the name bc its a little confusing w all of the local vars

packages/frontend-v2/utils/plan/updatePlanOnDragEnd.ts
```diff
@@ -29,82 +29,95 @@ import { isCourseInTerm } from "./isCourseInTerm";
...
+        );
       }
 
-      // remove the class from the old term and add it to the new term
-      oldTerm.classes = oldTerm.classes.filter(
-        (course) => course.id !== draggedCourse.id
+      const draggedCourseDetails: ScheduleCourse2<unknown> =
+        draggedCourse.data.current.course;
+
+      const oldTerm = scheduleTerms.find((term) =>
```
jia614, line 67: this bit looks like the code at my above comment-- i think u can just move it to the top to avoid recalculating the old term, but if it messes up the logic then j ignore

jia614 (approved):
looks gud to merge after merge conflicts r resolved + few nits. updatePlanOnDragEnd is starting to get real chunky & hard to read so it might be worth to move logic blocks into helpers, but up to u

PR #543, opened 2023-02-10
Dylan jia improve search fetching

dkd2101:
# Description

Created Hooks for the the search functions the Search API. 

Closes # (518)

## Type of change

- [ X] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

Manually tested that the behavior of the search component and the sidebar was the same.

# Checklist:


packages/frontend-v2/components/AddCourseModal/AddCourseModal.tsx
```diff
@@ -41,30 +40,16 @@ export const AddCourseModal: React.FC<AddCourseModalProps> = ({
...
-        handleApiClientError(error as Error, router);
-      }
-      setIsLoading(false);
-    }
-  };
+  const {
+    courses,
+    isLoading: isCoursesLoading,
+    error,
+  } = useSearchCourses(searchQuery);
```
aryanshah701, line 52: Ahh didn't realize you could simply just pass in the state into the hook without having to do anything we discussed haha. Good call!!

packages/frontend-v2/components/Sidebar/Sidebar.tsx
```diff
@@ -115,79 +113,34 @@ const Sidebar: React.FC<SidebarProps> = memo(({ selectedPlan }) => {
...
-      return;
-    }
+  const majorCourses = getAllCoursesInMajor(major, concentration);
 
-    const concentrationRequirements: IRequiredCourse[] = [];
-    getRequiredCourses(
-      concentration?.requirements ?? [],
-      concentrationRequirements
-    );
+  const { courses, isLoading: isCoursesLoading } =
```
aryanshah701, line 138: error state here is very important: essentially handling the state where search doesn't return any courses.

packages/frontend-v2/hooks/useFetchSearchCourse.ts
```diff
@@ -0,0 +1,61 @@
...
+type FetchCourseReturn = FetchCourseResponse & {
+  course?: ScheduleCourse2<null> | null;
+  isLoading: boolean;
+};
+
+/**
+ * @param subject - The type of class that we are fetching from SearchNEU
+ * @param classId - The identification number of the class as a string
+ */
+export function useFetchSearchCourse(
```
aryanshah701, line 20: let's remove "search" from all our hooks, it's kinda confusing(like are we searching with this hook?) and besides, fetching from SearchNEU using the SearchClient is an implementation detail of the hook that isn't very useful. 

When using the hooks, all we care about is the returned data(courses). How we get them could evolve. Never know, down the road we may not even fetch from SearchNEU 🤯

packages/frontend-v2/hooks/useFetchSearchCourse.ts
```diff
@@ -0,0 +1,61 @@
...
+}
+
+export const fetchSearchCourse = async (
+  subject: string,
+  classId: string
+): Promise<ScheduleCourse2<null> | null> => {
+  const course = await SearchAPI.fetchCourse(constructKey(subject, classId));
+
+  return course;
+};
```
aryanshah701, line 45: nit: helper not needed, doesn't the same thing as SearchAPI.fetchCourse

packages/frontend-v2/utils/plan/getAllCoursesInMajor.ts
```diff
@@ -0,0 +1,62 @@
+import {
```
aryanshah701, line 1: awesome job with this!

packages/frontend-v2/hooks/useFetchSearchCourse.ts
```diff
@@ -0,0 +1,61 @@
+import useSWR, { SWRResponse } from "swr";
```
aryanshah701, line 1: I don't think we use this hook anywhere, since we only fetch 1 course in the getRequiredCourseCoreqs function. And over there we use the fetching function directly.

If yall can make a good argument to keep it I'm down, otherwise might be nice to reduce code.

packages/frontend-v2/utils/course/getCourseCoreqs.ts
```diff
@@ -19,7 +15,7 @@ export async function getRequiredCourseCoreqs(
     await Promise.all(
       coreqCourses.map(async (coreqCourse: INEUReq) => {
         if (isINEUPrereqCourse(coreqCourse)) {
-          const course = await SearchAPI.fetchCourse(
+          const course = await fetchSearchCourse(
```
aryanshah701, line 18: For now, we should be able to just use SearchAPI.fetchCourse after the SearchAPI is updated to not take in queries. 

Down the line, we need to think about using both the cache and batch API. Added an issue for that: #546

aryanshah701 (changes requested):
Awesome job yall!! This was a very very involved task and you guys tackled it like champs! Some very impressive work here that is so critical and was delivered very quickly :CHUNGUS:

I know there's a lot of comments, but that's just a result of this task having many design decisions + the PR being so large. Take your time to go through them!

Finally, once done, could yall test and take vids/images of the following:

- Search courses: loading -> success state. then loading -> error state.
- Sidebar: loading -> success state. then loading -> error state.

If the above stuff looks good, then this ticket is a MASSIVE LEAP for us! Also, for error state, you could just temporarily simulate an error state in the fetching function in the SearchClient.

packages/frontend-v2/hooks/useFetchSearchCourses.ts
```diff
@@ -0,0 +1,50 @@
...
+    courses? : ScheduleCourse2<null>[];
+    isLoading : boolean;
+}
+
+/**
+ * 
+ * @param subject - the type of class that we are fetching from SearchNEU
+ * @param classId - the identification number of the class as a string
+ */
+export function useFetchSearchCourses(courses: {subject: string, classId: string}[]): FetchCoursesReturn{
```
BrandonLim8890, line 18: Super nit: I would just type this out separately like `type course = {...}` and then `courses: course[]`

packages/frontend-v2/hooks/useSearchCourses.ts
```diff
@@ -0,0 +1,67 @@
...
+>;
+
+type SearchCoursesReturn = SearchCoursesResponse & {
+  courses?: ScheduleCourse2<null>[];
+  isLoading: boolean;
+  mutateCourses: KeyedMutator<ScheduleCourse2<null>[]>;
+};
+
+/**
+ * @param subject - The type of class that we are fetching from SearchNEU
```
BrandonLim8890, line 18: update argument documentation

packages/frontend-v2/components/AddCourseModal/AddCourseModal.tsx
```diff
@@ -123,28 +104,31 @@ export const AddCourseModal: React.FC<AddCourseModalProps> = ({
...
-                key={getCourseDisplayString(searchResult)}
-                searchResult={searchResult}
-                addSelectedCourse={addSelectedCourse}
-                isResultAlreadyInTerm={isCourseInCurrTerm(searchResult)}
-                isResultAlreadySelected={isCourseAlreadySelected(searchResult)}
-              />
-            ))}
+            {error && (
+              <Text fontWeight="semibold" textAlign="center">
+                Error searching courses
```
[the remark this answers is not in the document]
jia614: i took a quick attempt at adding hyperlinks inside the tooltip with <span> and <a>, but it didn't seem to work at surface level so I just made the message a little more specific. I can look into it a little more but im guessing it shouldn't block merging this pr in

aryanshah701 (approved):
🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢

PR #558, opened 2023-02-26
Styling for removing courses when dragged outside the schedule

### LINKED ISSUE #538: Styling for Delete on Drag
Builds on #537. Just need to add conditional styling when hovering over drop to delete.

aryanshah701:
# Description

Issue with current implementation(Flickering):
- If the course is not over anything, we say it needs to be deleted(isRemove is true)
- However, for 1 render cycle, when the course is just picked up, over is null even though the course is over a droppable
- Hence, for 1 render cycle, we style the course as though it is being delete
- This leads to a flicker of the remove course styles

Fix:
- The only fix I could come up with was making the whole screen a droppable called delete area
- Droppables can be nested so this works well in our favor. The most nested droppable takes precedence, just like we would like.

Other changes:
- Remove "co-op" magic string from codebase
- Add comments detailing whatever I learnt while doing this ticket
- ScheduleCourse return was becoming massive, so moved some code out into sub components

Callout:
- I didn't style this exactly like the figma. I would like to discuss this with designers cause the color change to blue when the course i dragged outside the plan is weird. Plus, the white cross doesn't seem "daunting" enough for something like a "delete" action.
- This is fine though, we have the right foundations to easily style the remove course state now.

Closes #538

## Type of change

# How Has This Been Tested?

Deployed this to staging for GM demo so just check out staging.

# Checklist:

### DIFF
--- a/packages/frontend-v2/components/ScheduleCourse/ScheduleCourse.tsx
+++ b/packages/frontend-v2/components/ScheduleCourse/ScheduleCourse.tsx
@@ -1,4 +1,4 @@
-import { CloseIcon, DeleteIcon } from "@chakra-ui/icons";
+import { DeleteIcon } from "@chakra-ui/icons";
 import { Flex } from "@chakra-ui/react";
 import { useDraggable } from "@dnd-kit/core";
 import { CSS } from "@dnd-kit/utilities";
@@ -8,7 +8,7 @@ import {
   ScheduleCourse2,
 } from "@graduate/common";
 import { forwardRef, useEffect, useState } from "react";
-import { DELETE_COURSE_AREA_DND_ID } from "../../pages/home";
+import { DELETE_COURSE_AREA_DND_ID, isCourseFromSidebar } from "../../utils";
 import { ReqErrorModal } from "../Plan/ReqErrorModal";
 import { COOP_BLOCK } from "../Sidebar";
 
@@ -20,11 +20,11 @@ interface DraggableScheduleCourseProps {
   removeCourse?: (course: ScheduleCourse2<unknown>) => void;
   isEditable?: boolean;
   isDisabled?: boolean;
-  isFromSidebar?: boolean;
   /** Only provide this prop to the overlay course being dragged around the screen. */
   setIsRemove?: (val: boolean) => void;
 }
 
+/** This is the static course on the page that can be dragged around. */
 export const DraggableScheduleCourse: React.FC<
   DraggableScheduleCourseProps
 > = ({
@@ -33,7 +33,6 @@ export const DraggableScheduleCourse: React.FC<
   preReqErr = undefined,
   coReqErr = undefined,
   isEditable = false,
-  isFromSidebar = false,
   isDisabled = false,
   setIsRemove,
 }) => {
@@ -41,7 +40,6 @@ export const DraggableScheduleCourse: React.FC<
     useDraggable({
       id: scheduleCourse.id,
       data: {
-        isFromSidebar: isFromSidebar,
         course: scheduleCourse,
       },
       disabled: isDisabled,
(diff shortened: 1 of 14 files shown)


packages/frontend-v2/components/ScheduleCourse/ScheduleCourse.tsx
```diff
@@ -133,42 +148,12 @@ export const ScheduleCourse = forwardRef<
...
-              d="M3,2 C2.44771525,2 2,1.55228475 2,1 C2,0.44771525 2.44771525,0 3,0 C3.55228475,0 4,0.44771525 4,1 C4,1.55228475 3.55228475,2 3,2 Z M3,6 C2.44771525,6 2,5.55228475 2,5 C2,4.44771525 2.44771525,4 3,4 C3.55228475,4 4,4.44771525 4,5 C4,5.55228475 3.55228475,6 3,6 Z M3,10 C2.44771525,10 2,9.55228475 2,9 C2,8.44771525 2.44771525,8 3,8 C3.55228475,8 4,8.44771525 4,9 C4,9.55228475 3.55228475,10 3,10 Z M7,2 C6.44771525,2 6,1.55228475 6,1 C6,0.44771525 6.44771525,0 7,0 C7.55228475,0 8,0.44771525 8,1 C8,1.55228475 7.55228475,2 7,2 Z M7,6 C6.44771525,6 6,5.55228475 6,5 C6,4.44771525 6.44771525,4 7,4 C7.55228475,4 8,4.44771525 8,5 C8,5.55228475 7.55228475,6 7,6 Z M7,10 C6.44771525,10 6,9.55228475 6,9 C6,8.44771525 6.44771525,8 7,8 C7.55228475,8 8,8.44771525 8,9 C8,9.55228475 7.55228475,10 7,10 Z"
-            ></path>
-          </svg>
-          <p style={{ fontWeight: "bold"}}>
-            {`${courseToString(scheduleCourse)} `}
-            <span style={{ marginLeft: "2px", fontWeight: "normal"}}>
-              {scheduleCourse.name}
-            </span>
-          </p>
-        </div>
```
aryanshah701, line 171: simply moved this into its own components

packages/frontend-v2/components/Sidebar/Sidebar.tsx
```diff
@@ -18,7 +18,11 @@ import { handleApiClientError } from "../../utils";
 import axios from "axios";
 import { useRouter } from "next/router";
 import { useMajor } from "../../hooks/useMajor";
-import { WorkerMessage, WorkerMessageType, WorkerPostInfo } from "../../validation-worker/worker-messages";
```
aryanshah701, line 21: this file has a lotta prettier changes. only actual change is adding the COOP_BLOCK constant course so that we don't have magic strings in the codebase

packages/frontend-v2/pages/home.tsx
```diff
@@ -132,7 +142,7 @@ const HomePage: NextPage = () => {
    */
   const handleDragStart = (event: DragStartEvent) => {
     const { active } = event;
-    setActiveCourse(active.data.current?.course);
+    setActiveCourse({ ...active.data.current?.course, id: active.id });
```
aryanshah701, line 145: adding the id in so that we can check if a course is from the sidebar or not

packages/frontend-v2/pages/home.tsx
```diff
@@ -251,6 +261,7 @@ const HomePage: NextPage = () => {
             coReqErr={undefined}
             preReqErr={undefined}
             isRemove={isRemove}
+            isFromSidebar={activeCourse.id.startsWith(SIDEBAR_DND_ID_PREFIX)}
```
aryanshah701, line 264: this is important so that we don't style courses from the sidebar with the "delete course" styles

packages/frontend-v2/utils/theme/index.ts
```diff
@@ -105,12 +105,12 @@ const space = {
   "2xl": "3rem",
   "3xl": "4rem",
   "4xl": "5rem",
-  "5xl": "6rem"
+  "5xl": "6rem",
```
aryanshah701, line 108: just prettier changes here

packages/frontend-v2/components/ScheduleCourse/ScheduleCourse.tsx
```diff
@@ -46,7 +48,7 @@ export const DraggableScheduleCourse: React.FC<
     });
 
   useEffect(() => {
-    if (setIsRemove) setIsRemove(over === null);
+    if (setIsRemove) setIsRemove(over?.id === DELETE_COURSE_AREA_DND_ID);
```
BrandonLim8890, line 51: I feel like the overlap between the overlay course being dragged and the original course is really confusing for anyone trying to understand the logic behind draggable course. Might be worth seperating them into different components with different props, but have the same styling etc.

BrandonLim8890 (approved):
LGTM! I think the only thing is just to abstract scheudle course and the overlay to another component with different props but same underlying styling/components. But I think this can be shifted to another issue.

packages/frontend-v2/components/ScheduleCourse/ScheduleCourse.tsx
BrandonLim8890: ALSO, we can actually replace this state with `useMemo` instead. Its like a glorified `useEffect` designed to udpate a value that relies on another piece of state. I think we can start implementing this more and more in our codebase.

aryanshah701: I agree on the overlap, let me spend some time refactoring!

packages/frontend-v2/components/ScheduleCourse/ScheduleCourse.tsx
```diff
@@ -62,24 +64,30 @@ export const DraggableScheduleCourse: React.FC<
       attributes={attributes}
       transform={CSS.Translate.toString(transform)}
       isDisabled={isDisabled}
+      isFromSidebar={isFromSidebar}
```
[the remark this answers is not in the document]
aryanshah701: ahhh that's true, thank you!!

PR #567, opened 2023-03-10
Revamp signup and login UX

### LINKED ISSUE #552: Revamp signup/login pages according to new designs
Implement the new designs for signup and login pages from the figma handoff.

aryanshah701:
# Description

https://user-images.githubusercontent.com/30478978/224428795-405bc5e3-b56e-4865-b636-bedb2a1d2af1.mov

Update signup and login pages to reflect new designs.

Some notable changes in the process of doing so:
- Update headers to be consistent through a common pre and post auth header: this comes from alpha feedback
- Link components that use NextLink
- Debug logs for when user isn't authorized cause email isn't confirmed
- Handle email already exists case for signup

Closes #552 

## Type of change

# How Has This Been Tested?

Locally, check out video.

# Checklist:


packages/common/src/api-dtos.ts
```diff
@@ -63,6 +63,14 @@ export class UpdatePlanDto {
 }
 
 export class SignUpStudentDto {
+  @IsOptional()
```
BrandonLim8890, line 66: Do we want to include this here or is this just to fit the current designs and will be later removed in a future PR?

packages/frontend-v2/pages/login.tsx
```diff
@@ -61,92 +38,65 @@ const LoginForm = () => {
...
-        as="b"
-        mb="xl"
-        textAlign="center"
-      >
-        Hey there!
-      </Text>
+    <AuthForm
+      onSubmit={handleSubmit(onSubmitHandler)}
+      headingText="Welcome Back!"
+      inputs={
```
BrandonLim8890, line 55: Instead of passing in React Nodes as props, would it be better to pass in the inputs as children?

packages/frontend-v2/pages/signup.tsx
```diff
@@ -64,97 +39,90 @@ const SignUpForm = () => {
...
+      onSubmit={handleSubmit(onSubmitHandler)}
+      headingText="Create an Account"
+      inputs={
+        <>
+          <Flex columnGap="md">
+            <GraduateInput
+              type="text"
+              id="firstName"
+              placeholder="First Name"
+              {...register("firstName")}
```
BrandonLim8890, line 63: So this won't work because react hook form will simply create a field `firstName` in the payload and send that to the BE when it should be `fullName` which includes both `firstName` and `lastName`. We need to pull out both of these values on submit and then combine them together into `fullName` because right now, the post request is not setting `fullName` in the BE. 

Another approach we can take is to make `firstName` and `lastName` distinct in the BE, which I think is better because it can be difficult to guess first and last name from just `fullName`. Someone with the name "Yu Ping Chao" would have "Yu Ping" as their first name, but would be difficult to identify if we just stored `fullName: "Yu Ping Chao"` in the DB.

packages/frontend-v2/pages/signup.tsx
```diff
@@ -64,97 +39,90 @@ const SignUpForm = () => {
...
+              helpMessage="Optional"
+            />
+            <GraduateInput
+              type="text"
+              id="lastName"
+              placeholder="Last Name"
+              {...register("lastName")}
```
BrandonLim8890, line 70: Connected to the comment above ^^.

BrandonLim8890 (approved):
LGTM! Just one comment detailing how registering with last name and first name defined doesn't actually define them in the BE due to the discrepancy in the payload. We can, however, probably adjust this in another ticket as it affects #554 as well as @Nickdaco and I are finding it hard to extract `firstName` and `lastName` from `fullName`.

packages/frontend-v2/pages/login.tsx
aryanshah701: a solid case could be made for both! passing in as inputs makes it more meaningful here since this component isn't some layout or scaffolding but rather meant to render the entire form given the input fields. this wouldn't be clear if we rendered the children as the inputs of the form.

packages/frontend-v2/pages/signup.tsx
aryanshah701: ahhh good catch, will update this!

packages/common/src/api-dtos.ts
aryanshah701: didn't realize the student entity had a fullname and not a first and last name... i'll update this to be an optional full name

PR #571, opened 2023-03-12
Optional Email Confirmation

### LINKED ISSUE #564: Make email confirmation optional
New user flow:
- Sign up -> email confirmation page(link to home on this page so they can do it later if they want) -> home.

When we fetch a user we should know if the user has their email confirmed. If the user's email isn't confirmed, for now display a warning model the first time they visit home asking them to confirm their email with the resend link in case the previous is expired. The warning model should communicate that they cannot recover their account if they forget their password and haven't confirmed their email.

We can also show a warning symbol in the user profile when we implement view/edit user details.

BrandonLim8890:
# Description

Instead of forcing the user to confirm their email, we now allow the user to bypass this and use the homepage regardless of if they have confirmed their email.

Closes #564 

## Type of change

# How Has This Been Tested?

Ran through workflows

https://user-images.githubusercontent.com/20124104/224527230-cd20d56f-18d2-4a43-a55c-41c51ff26c4d.mov

# Checklist:

### DIFF
--- a/packages/frontend-v2/components/Error/ConfirmEmailWarningModal.tsx
+++ b/packages/frontend-v2/components/Error/ConfirmEmailWarningModal.tsx
@@ -1,0 +1,74 @@
+import {
+  Modal,
+  ModalBody,
+  ModalHeader,
+  useDisclosure,
+  Text,
+  ModalFooter,
+  Button,
+  Link,
+  ModalContent,
+  ModalOverlay,
+} from "@chakra-ui/react";
+import { API } from "@graduate/api-client";
+import { StudentModel } from "@graduate/common";
+import { useLocalStorage } from "../../hooks";
+import { toast } from "../../utils";
+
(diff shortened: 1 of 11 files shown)


aryanshah701 (approved):
looks goood! Lemme rebase this on top of my changes, I'll merge :))

packages/frontend-v2/components/Error/ConfirmEmailWarningModal.tsx
```diff
@@ -0,0 +1,74 @@
+import {
```
BrandonLim8890, line 1: The current implementation of this modal uses localStorage to find out if it is the user's first time visiting this page with the `alreadyVisited` key. However, because next loads this page before localStorage can be accessed, the default value is null, and so the component thinks that the user HAS visited the page before, and thus the warning modal pops up.

PR #610, opened 2023-04-12
Hotfix for multiple plans

AlpacaFur:
# Description

Somehow the updatedAt property is being sent as a string somewhere along the process so when `getTime` is called, it crashes the app. To prevent this, I'm just converting the dates explicitly. This has no effect on existing dates (so it's safe to include even if the issue isn't consistent), but converts strings to the correct date, allowing the `getTime` method to be run.

Something is wrong somewhere else in the app that's causing this, but this mitigates the symptoms in the short term.

## Type of change

# How Has This Been Tested?

Manually tested to check that this fixes the problem.

# Checklist:


aryanshah701 (approved):
<3

PR #636, opened 2023-10-22
UI updates

AngelaZQ1:
# Description

Made the following UI modifications:
- lighten sidebar bg color (will match the sidebar in the new add course modal) and added a right border
- lighten bg color of `ScheduleTerm`
- update "Add Course" button colors
- decrease blue button and delete plan button border to .5px
- decrease spacing between input label and input box
- change `lg` input sizes to `md` for consistency
- moved the "no major" checkbox to the left of the text
- added a + icon to the add year button
- removed an extra '"' from the error modal
- italicized the `SectionRequirement` text 
- decrease spacing between "graduate" and "your way" on home page

Also updated the `primary.neutral` colors in the theme folder to remove unused grays and replace some with the names used in figma (gray0, gray, gray2, gray3). `primary.neutral.main` was replaced with `primary.neutral.gray`.

Old:
Screenshot 2023-10-22 at 2 39 16 PM

New:
Screenshot 2023-10-22 at 2 39 02 PM

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

# Checklist:

- [X] Any dependent changes have been merged and published in downstream modules


packages/frontend-v2/utils/theme/index.ts
```diff
@@ -49,15 +49,13 @@ const colors = {
   },
   neutral: {
     main: "#e7ebf1",
```
daisykucharski, line 51: [nit] maybe we can condense the 2 instances of #e7ebf1 into one instance in this file?

packages/frontend-v2/components/Plan/TransferCourses.tsx
```diff
@@ -71,7 +71,7 @@ export const TransferCourses: React.FC<TransferCoursesToggleProps> = ({
   };
 
   return (
-    <Flex direction="column" backgroundColor={"neutral.main"}>
+    <Flex direction="column" backgroundColor={"#e7ebf1"}>
```
daisykucharski, line 74: [nit] this color is defined in the theme file, we should probably use that for

daisykucharski (approved):
LGTM minus a few nits

packages/frontend-v2/utils/theme/index.ts
AngelaZQ1: so true i missed that

packages/frontend-v2/components/Header/AccountOverview.tsx
```diff
@@ -130,7 +130,7 @@ export const AccountOverview: React.FC<AccountOverviewProps> = ({
                 />
               </Flex>
               <FormControl>
-                <FormLabel>
+                <FormLabel mb="1px">
```
BrandonLim8890, line 133: Same comment here for sizing

packages/frontend-v2/utils/theme/index.ts
```diff
@@ -48,16 +48,13 @@ const colors = {
     },
   },
   neutral: {
-    main: "#e7ebf1",
-    50: "#eff2f6",
-    100: "#d2d7de",
-    200: "#b3bcc8",
-    300: "#f1f5fa",
+    gray0: "#f4f6f9",
```
BrandonLim8890, line 51: Can we get different names for this? Either create a new type under neutral and then add weights to them. Eg instead of `neutral.gray0` it would be `neutral.gray.100 neutral.gray.200` etc

BrandonLim8890 (changes requested):
CHUNGUS CHANGES WE LOVE A DESIGNER THAT CAN CODE. Just some small nits here and there but the biggest thing is to create a "theme" with the styling we write. So instead of specifying `1px` for borders or for spacing in multiple components, we can just define it within a file and call `3xs` for example. This standardizes our code, so that when we get designs from you guys, we can even just say that a border/padding/spacing is simply `s, m, l` etc. You can add to this file [here](https://github.com/sandboxnu/graduatenu/blob/main/packages/frontend-v2/utils/theme/index.ts).

packages/frontend-v2/utils/theme/index.ts
```diff
@@ -52,10 +52,7 @@ const colors = {
     gray: "#e7ebf1",
     gray2: "#d2d8e3",
     gray3: "#b4bbc8",
-    300: "#f4f6f9",
-    400: "#7586a0",
-    700: "#e2e6ec",
-    900: "#d2d8e2",
```
AngelaZQ1, line 61: 300 and 700 arent used, 900 is basically gray2, and changed 400 to gray4

packages/frontend-v2/utils/theme/index.ts
```diff
@@ -48,17 +48,11 @@ const colors = {
...
-    100: "#d2d7de",
-    200: "#b3bcc8",
-    300: "#f1f5fa",
-    400: "#7586a0",
-    500: "#e7ebf1",
-    600: "#485569",
-    700: "#e2e6ec",
-    800: "#1f242c",
-    900: "#d2d8e2",
+    gray0: "#f4f6f9",
```
BrandonLim8890, line 51: Can you just change these to `gray.0`. So like gray would be its own object. This way we aren't doing `gray0`, `gray1`. It would instead be `gray.50`, or `gray.100`

BrandonLim8890 (approved):
Just a small nit, once you change it you can merge! 💯

packages/frontend-v2/utils/theme/index.ts
AngelaZQ1: Hhhh gotcha i see what you mean now

AngelaZQ1: Would it be fine to just do `neutral.100` instead of `neutral.gray.100` since neutral is just gonna be for grays?

BrandonLim8890: Yea if we feel like there wouldn't be anymore neutrals, then this works

PR #646, opened 2023-10-29
Add "passwordConfirm" to Logging Denylist

AlpacaFur:
# Description
We are currently logging the passwordConfirm field production which is bad security practice. This PR removes the field by adding it to the denylist.

## Type of change

# How Has This Been Tested?
Tested locally.


rael346 (commented):
lol approved

BrandonLim8890 (approved):

PR #664, opened 2023-11-18
feat: allow temp vercel URL as origin (CORS)

lukejianu:
# Description

Add a temporary Vercel URL to the allowlist for the backend, so that we can test the new setup (Vercel frontend) in staging.


Christinarlong (approved):
wow lgtm

PR #681, opened 2023-12-03
Fix bug with adding a transfer course without a fullname

### LINKED ISSUE #680: Unable to add transfer course as a user with an empty name
Bug where we can't add a transfer course because the BE validation is angry at us when we try to update a student with an empty name. Just need to change the DTO to remove the check for empty full name

BrandonLim8890:
# Description

Previously, we would check if the user has a fullname when updating. When we made the change to allow users to not have a fullname, it would error in our backend. This PR just removes that check.
This means that previously, a user with an empty fullname could not add a new transfer course.

Closes #680

## Type of change

# How Has This Been Tested?

Users without fullname can now add transfer courses.

# Checklist:

### DIFF
--- a/packages/common/src/api-dtos.ts
+++ b/packages/common/src/api-dtos.ts
@@ -84,7 +84,6 @@ export class SignUpStudentDto {
 }
 export class UpdateStudentDto {
   @IsOptional()
-  @IsNotEmpty()
   @IsString()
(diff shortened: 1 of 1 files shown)


DylanCerenov (approved):
LGTM 👍

PR #729, opened 2024-03-28
Beta major support

### LINKED ISSUE #666: Supporting Beta Majors
Our scrapers are able to scrape the majors that exist within NEU, but the scraper may not accurately scrape all information about a major. As such, to have a valid major we need to manually verify each of these majors, which can be very time consuming. As a compromise, we can add support for "Beta majors" which are majors that have not been manually verified. This would mean that we require a warning modal that explicitly states that these majors have not been validated, and will likely have errors.

https://www.figma.com/file/yUflVFVxjMynm7gGymLpQY/%F0%9F%8F%97-Current-Work?type=design&node-id=4373-36187&mode=design&t=IKBrY1mvM8RaU9JK-0

KobeZ123:
# Description

Implement major metadata for check for verified majors
- update supportedMajors endpoint to add verified key 
- update select component to have different label and value for adding beta marking on major names 
- adding badge for beta major on sidebar 
- add checkbox for agreeing to use beta major 

Closes #666

## Type of change

# How Has This Been Tested?

- [X] Test both the add plan and edit plan modals for beta majors in select elements and checkbox for agreeing to beta major
- [X] Check the sidebar to ensure majors have badge to indicate beta majors

# Checklist:

### DIFF
--- a/packages/common/src/types.ts
+++ b/packages/common/src/types.ts
@@ -505,3 +505,9 @@ export interface MetaInfo {
   build_timestamp: Maybe<number>;
   environment: Maybe<string>;
 }
+
+// option object type for react-select
+export type OptionObject = {
+  label: string | number;
+  value: string;
+};
--- a/packages/common/src/types.ts
+++ b/packages/common/src/types.ts
@@ -143,6 +143,18 @@ export interface Major2 {
   totalCreditsRequired: number;
   yearVersion: number;
   concentrations?: Concentrations2;
+  metadata?: MajorMetadata;
+}
+
+/**
+ * Metadata for a major.
+ *
+ * @param verified   Whether the major has been manually verified.
+ * @param lastEdited The last time the major was edited MM/DD/YYYY.
+ */
+export interface MajorMetadata {
+  verified: boolean;
+  lastEdited: string;
 }
 
 /**
@@ -456,9 +468,10 @@ export interface IScheduleCourse {
 export type SupportedConcentrations = {
   concentrations: string[];
   minRequiredConcentrations: number;
+  verified: boolean;
 };
 
-// { majorName => { concentration, minRequiredConcentrations } }
+// { majorName => { concentration, minRequiredConcentrations, verified} }
 export type SupportedMajorsForYear = Record<string, SupportedConcentrations>;
 
 // { year => supported majors }
--- a/packages/common/src/types.ts
+++ b/packages/common/src/types.ts
@@ -509,5 +509,5 @@ export interface MetaInfo {
 // option object type for react-select
 export type OptionObject = {
   label: string | number;
-  value: string;
+  value: string | number;
 };
--- a/packages/frontend/components/Form/Select.tsx
+++ b/packages/frontend/components/Form/Select.tsx
@@ -4,6 +4,7 @@ import {
   FormErrorMessage,
   FormHelperText,
 } from "@chakra-ui/react";
+import { OptionObject } from "@graduate/common";
 import Fuse from "fuse.js";
 import { Control, FieldError, useController } from "react-hook-form";
 import Select from "react-select";
@@ -14,6 +15,11 @@ type PlanSelectProps = {
   label: string;
   helperText?: string;
   options: (string | number)[];
+  /**
+   * Optional list of (label, value) if the select option label and value should
+   * be different
+   */
+  optionObjects?: OptionObject[];
   /** Any side effects as a result of this field changing. */
   onChangeSideEffect?: (val: string | null) => void;
   /** The name of the react hook form field. */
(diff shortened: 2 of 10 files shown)


packages/common/src/types.ts
```diff
@@ -456,9 +468,10 @@ export interface IScheduleCourse {
 export type SupportedConcentrations = {
   concentrations: string[];
   minRequiredConcentrations: number;
+  verified: boolean;
```
rael346, line 471: This might be just me not being familiar with Graduate codebase after a while, but why is the `verified` property here necessary for concentration when a major already indicates that? (If a major is `verified`, that should mean the whole major, including its concentration, is verified right?)

packages/common/src/types.ts
```diff
@@ -143,6 +143,18 @@ export interface Major2 {
   totalCreditsRequired: number;
   yearVersion: number;
   concentrations?: Concentrations2;
+  metadata?: MajorMetadata;
```
rael346, line 146: I think we can make this metadata a requirement for every major since `verified` and `lastEdited` will always be in a major anyway. Thought for now I guess we can leave it as optional so it doesn't break the current majors. (We can make it not optional in a future ticket when we add new majors) @AlpacaFur what do you think?

rael346 (commented):
Just passing by since I'm in charge of scraper at the moment. I only reviewed the types for the major object, everything else I will leave to the actual reviewers since they know more about Graduate codebase than I do.

packages/frontend/components/Sidebar/Sidebar.tsx
```diff
@@ -226,6 +227,7 @@ const Sidebar: React.FC<SidebarProps> = memo(
         creditsToTake={major.totalCreditsRequired}
         helperText={sidebarHelperText}
         renderCoopBlock
+        renderBetaMajorBlock={major.metadata?.verified !== true}
```
BrandonLim8890, line 230: Again will remove `?` here in my follow up PR

packages/common/src/types.ts
BrandonLim8890: yea I plan to change this in a follow up PR when I add metadata to all current PRs

BrandonLim8890 (changes requested):
Looks good so far! Just some small suggestions

packages/frontend/components/Form/Select.tsx
```diff
@@ -35,6 +38,7 @@ type PlanSelectProps = {
 export const PlanSelect: React.FC<PlanSelectProps> = ({
   label,
   options,
+  optionObjects,
```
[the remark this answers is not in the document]
KobeZ123: yeah, that makes a lot of sense to me!

packages/frontend/components/Plan/AddPlanModal.tsx
```diff
@@ -296,6 +306,20 @@ export const AddPlanModal: React.FC<AddPlanModalProps> = ({
                         }}
                       />
                     )}
+                    {majorName && !isValidatedMajor && (
```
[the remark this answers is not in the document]
KobeZ123: If a major name is not selected, the checkbox will show up by default because isValidatedMajor will default to false. I assume we only want to show the checkbox if the selected major is a beta major.

packages/common/src/types.ts
KobeZ123: The list of majors endpoint for the add plan and edit plan modals use some kind of Record mapping relevant major information. Since we want to mark majors with a [BETA] tag in the modal, we also want to pass along the flag from the metadata if the major is verified. I guess technically the name of this type can be updated to SupportedMajorInformation, including concentrations and now this verified tag.

packages/frontend/components/Plan/AddPlanModal.tsx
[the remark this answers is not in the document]
BrandonLim8890: Can we leave a quick comment about this above this line as i don't think its that obvious

packages/frontend/components/Plan/AddPlanModal.tsx
```diff
@@ -34,6 +35,7 @@ import {
   cleanDndIdsFromStudent,
   createEmptySchedule,
   extractSupportedMajorNames,
```
BrandonLim8890, line 37: Nit: can we fix this error

packages/frontend/components/Plan/EditPlanModal.tsx
```diff
@@ -28,6 +32,7 @@ import {
 } from "../../hooks";
 import {
   extractSupportedMajorNames,
```
BrandonLim8890, line 34: nit: remove unneeded var here too

BrandonLim8890 (approved):
LGTM! just 2 nits about unused vars before you merge!

PR #794, opened 2025-01-20
Sidebar validation infinite loading fix

denniwang:
# Description

https://github.com/user-attachments/assets/5e0e7fa6-ee68-40af-bf4f-a450143c4307

Also added a `wdev` script 

Closes #617 Validation Loading Indefinetly
## Type of change

Please tick the boxes that best match your changes.

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update
- [ ] This requires a run of `yarn install`
- [ ] This has migration changes and requires a run of `yarn dev:migration:run`

# How Has This Been Tested?

Check screen recording

# Checklist:

- [ ] I have run the production builds in docker for the frontend/backend and ensure things run fine. Check README of repo on how to run if not sure.
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code where needed
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] I've run the end to end tests
- [ ] Any dependent changes have been merged and published in downstream modules

PR #805, opened 2025-02-05
Range update (fixed infinite loading)

denniwang:
# Description

In the major validator, we have a set of courses that include all the REQUIRED courses that SHOULD NOT be considered in the range validator. This fixed the infinite loading sidebar validators. I dont think this is optimal and would be inefficient because we are fetching and grabbing the required courses every time we want to validate. 

Closes #794 

## Type of change

# How Has This Been Tested?

Tested changes with debug statements to check what courses are excluded and which ones are being considered. 

# Checklist:

### DIFF
--- a/packages/common/src/major2-validation.ts
+++ b/packages/common/src/major2-validation.ts
@@ -354,8 +354,6 @@ export function getNecessaryCourses(
   for (const req of requirements) {
     crawlRequirement(req, tracker, requiredCourses);
   }
-  console.log("KOBE");
-  console.log(requiredCourses);
   return requiredCourses;
 }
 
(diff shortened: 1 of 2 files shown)


packages/common/src/major2-validation.ts
```diff
@@ -323,6 +341,50 @@ export function validateMajor2(
...
+}
+
+/** Crawls through the requirements, producing a list of necessary courses. */
+function crawlRequirement(
+  req: Requirement2,
+  tracker: CourseValidationTracker,
+  requiredCourses: Set<string>
+): void {
+  switch (req.type) {
+    // base cases
```
liamcarvajal, line 367: If you're keeping these comments, maybe they should be a little more formal. Otherwise looks good

packages/common/src/major2-validation.ts
```diff
@@ -193,13 +193,19 @@ interface CourseValidationTracker {
...
+
+  getNecessaryCourses(): Set<string>;
 }
 
 // exported for testing
 export class Major2ValidationTracker implements CourseValidationTracker {
   // maps courseString => [course instance, # of times taken]
   private currentCourses: Map<string, [ScheduleCourse2<unknown>, number]>;
 
+  private necessaryCourses: Set<string> = new Set();
```
KobeZ123, line 207: we can add some comments for what this aims to do 
// list of course names tracking courses that are necessary in the schedule (does not qualify as course for range requirement)

KobeZ123 (approved):
LGTM!

PR #812, opened 2025-03-06
Setting up vercel analytics

denniwang:
# Description

Adding vercel analytics. Ready to merge if approved by Kobe. 
https://vercel.com/docs/analytics
It will be useful to find visitor and page view statistics.


KobeZ123 (approved):
solid!

Issue #752, opened 2024-09-22
[DESIGN] Updating pre-requisite and co-requisite designs


KobeZ123:
**Summary:**
Implement the revamped pre-requisite and co-requisite designs

- The pre/co-requisite designs are rendered recursively (I think). Look into this and figure out how to implement the new designs accordingly

**Design:**
https://www.figma.com/design/yUflVFVxjMynm7gGymLpQY/%F0%9F%8F%97-Current-Work?node-id=6269-37326

PR #814, opened 2025-03-14
#752 Req Modal Improvements

23langloisj:
# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

Added a plus icon to the right of every course in the prereq/coreq error modal. Unfortunately, the design is still the same I don't see a way the design that we proposed would be possible so that might need looked at again if we want to consider changing the design. Might be worth moving to a new ticket.

Closes # 752

# How Has This Been Tested?
Screenshot 2025-03-14 at 7 28 00 PM

Upon Kobe's styling changes new ss looks like:

Screenshot 2025-03-24 at 8 49 48 PM
and appears to behave as expected for more complex cases

1) Drag in Algo in first semester
2) add courses that are needed as prereqs/coreqs

### DIFF
--- a/packages/frontend/components/Plan/Plan.tsx
+++ b/packages/frontend/components/Plan/Plan.tsx
@@ -28,6 +28,7 @@ interface PlanProps {
 }
 
 export const TotalYearsContext = createContext<number | null>(null);
+export const PlanContext = createContext<PlanModel<string> | null>(null);
 export const Plan: React.FC<PlanProps> = ({
   plan,
   mutateStudentWithUpdatedPlan,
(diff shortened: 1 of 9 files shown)


packages/frontend/components/Plan/Plan.tsx
```diff
@@ -28,6 +28,7 @@ interface PlanProps {
 }
 
 export const TotalYearsContext = createContext<number | null>(null);
+export const PlanContext = createContext<PlanModel<string> | null>(null);
```
KobeZ123, line 31: i like this

packages/frontend/components/Plan/ReqErrorModal.tsx
```diff
@@ -165,29 +187,124 @@ export const ReqErrorModal: React.FC<ReqErrorModalProps> = ({
 interface ParseCourseProps {
   course?: INEUReqError;
   parent: boolean;
+  term?: ScheduleTerm2<string>;
+  originalCourse?: ScheduleCourse2<unknown>;
 }
 
-// Look through the course error until there are no more errors!
-// TODO: Fix the styling!
 const ParseCourse: React.FC<ParseCourseProps> = ({
```
KobeZ123, line 194: what is the current issue with this? is it because top-level requirements render differently from nested requirements?

23langloisj (commented):
this looks good to me, really clever way of styling the ParseCourse component. One thing to note is that it would be nice to be able to unselect a course in the modal, but that's not too important and would probably make a good new hire ticket for the start of next semester. this LGTM

denniwang:
why does the Preview deployment not connect to the backend?

KobeZ123 (approved):

PR #832, opened 2025-04-07
cleaner ssl typeorm fix

23langloisj:
# Added AWS RDS ssl certificate validation for the database

Updated SSL configuration for production and staging environment to get rid of the db error when running the env locally.


packages/api/assets/RDS.us-east-1.ca-bundle.pem
denniwang, line 1: should we make this private ?

KobeZ123 (approved):

PR #855, opened 2025-09-02
clean console when running backend tests

23langloisj:
Removing console logs from the collator so linting warnings are removed from console when we run backend tests. To run backend tests:

cd packages
cd api
yarn jest OR yarn test

All tests should pass

### DIFF
--- a/packages/api/src/major/major-collator.ts
+++ b/packages/api/src/major/major-collator.ts
@@ -178,18 +178,18 @@ async function collateMajors() {
     })
   );
 
-  console.log(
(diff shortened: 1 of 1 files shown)


denniwang (approved):
i think would be helpful to have it back in the future ngl but all g

23langloisj:
> i think would be helpful to have it back in the future ngl but all g

Yeah it's definitely useful information but for some reason jest gets mad that there are logs after the expects and completely clogs the console. If we end up putting testing on the backburner I say we can call it back or we can look more into why the console is clogging and have both. This is just the mindless fix for now

PR #866, opened 2025-09-17
Logging cleanup

denniwang:
# Description

hides the body of responses that have to do with /major and /minor because they take up the entire terminal space and dont help


23langloisj (approved):
Code successfully hides all the logs of the major and minor info and successfully removes clutter when running the app locally. lgtm

PR #868, opened 2025-09-17
2025 majors, 2023-2025 minors

denniwang:
# Description

added 2025 majors

todo:
- need to push major scraper changes to main
- log major-scraper branch to jsons and have ci check that imported majors must be from major-scraper/main
- have cli tool to check major diffs between years


23langloisj (commented):
My only concern is that we probably shouldn't roll out the 2025 majors without the 2025 minors as well. If we are able to scrap the minors I say we do it here and merge it all in together. If it's like 2023 and we are unable to get the minors then I'm fine with sending this how it is.

